Issue #9, opened 2026-09-05
Publishing workflow calls Gradle tasks that no longer exist after the maven-publish plugin upgrade


pravbeseda:
## Problem

`.github/workflows/publish.yml` runs two Gradle tasks that no longer exist after the
toolchain upgrade to `com.vanniktech:gradle-maven-publish-plugin:0.34.0`:

- `publishAllPublicationsToMavenCentralRepository` (`.github/workflows/publish.yml:25`)
- `closeAndReleaseRepository` (`.github/workflows/publish.yml:33`)

Verified against the build:

```
./gradlew :library:tasks --all | grep -i publish
publish
publishMavenPublicationToMavenLocal
publishToMavenLocal
```

Neither task is registered any more. Plugin 0.30+ replaced them with a single
`publishAndReleaseToMavenCentral`, configured through a `mavenPublishing { }` block rather
than through the `SONATYPE_HOST` Gradle property.

Related: `gradle.properties:23` still sets `SONATYPE_HOST=S01`. The OSSRH host it points at
was retired in favour of the Central Portal, so that property is stale regardless of the
plugin version.

## Steps to reproduce

1. Create a GitHub release, which triggers `publish.yml`.
2. The `Upload release` step fails with `Task 'publishAllPublicationsToMavenCentralRepository'
   not found in root project`.

## Impact

Releases to Maven Central cannot be published. Nothing else in the build is affected —
`./gradlew build` and the test suite are green.

## Suggested fix

Configure `mavenPublishing { publishToMavenCentral(); signAllPublications() }` in
`library/build.gradle`, drop `SONATYPE_HOST` from `gradle.properties`, and point
`publish.yml` at `publishAndReleaseToMavenCentral`. Needs Central Portal credentials in the
repository secrets instead of the old OSSRH pair.

PR #10, opened 2026-09-05
Upgrade toolchain to Gradle 8.13 / AGP 8.13 / Kotlin 2.1.10, fix the pre-commit hook

### LINKED ISSUE #8: pre-commit hook calls a non-existent 'ktlint' Gradle task and blocks commits
## Problem

The pre-commit hook installed by `./gradlew installGitHook` invokes a Gradle task that does not exist, so every commit made with the hook installed is rejected.

`scripts/pre-commit:8`

```bash
./gradlew :library:spotlessApply
./gradlew ktlint     # <-- no such task

RESULT=$?
...
[ $RESULT -ne 0 ] && exit 1
```

`ktlint` is not a Gradle task in this project. It appears only as a step inside the Spotless configuration (`spotless.gradle:12`):

```groovy
ktlint('0.33.0').userData(['indent_size': '4', 'continuation_indent_size': '4'])
```

No ktlint Gradle plugin is applied anywhere (`build.gradle`, `library/build.gradle`, `sample/build.gradle`). The Spotless plugin 4.0.0 registers `spotlessApply` / `spotlessCheck`, not `ktlint`. The failing invocation sets `RESULT` to a non-zero value and the hook exits 1, aborting the commit.

## Steps to reproduce

1. `./gradlew installGitHook`
2. Change any file and `git commit`
3. The hook fails on `./gradlew ktlint` and the commit is aborted.

## Impact

Limited today, because `installGitHook` is manual — the line that would wire it into the build is commented out at `build.gradle:37`:

```groovy
// tasks.getByPath(':app:preBuild').dependsOn installGitHook
```

Anyone who does run `installGitHook` cannot commit until they delete `.git/hooks/pre-commit`.

## Suggested fix

Replace the second command with the check task that actually exists, so the hook formats and then verifies:

```bash
./gradlew :library:spotlessApply
./gradlew :library:spotlessCheck
```

## Note on verification

This was found by reading the build scripts, not by running the hook: Gradle 7.5 (`gradle/wrapper/gradle-wrapper.properties`) refuses to start on the only JDK installed on the machine used here (JDK 21 — `Unsupported class file major version 65`), so `./gradlew tasks --all` could not be executed to confirm the task list empirically.

pravbeseda:
Moves the build off a toolchain that no longer starts on a current JDK, and fixes the
pre-commit hook along the way.

Closes #8.

## Why

Gradle 7.5 refuses to run on JDK 19+ (`Unsupported class file major version`), so on a
machine with only a recent JDK installed nothing in this repository could be built or
tested. Gradle 8.13 needs AGP 8, AGP 8 needs Kotlin 1.9+, and each of those pulls its own
required change — the whole chain is here.

## What changed

| Before | After |
| --- | --- |
| Gradle 7.5 | Gradle 8.13 |
| AGP 7.4.2 | AGP 8.13.0 |
| Kotlin 1.7.21 | Kotlin 2.1.10 |
| Spotless 4.0.0 + ktlint 0.33.0 | Spotless 8.9.0 + ktlint 1.8.0 |
| Mockito 4.7.0 + mockito-inline | Mockito 5.20.0 |
| maven-publish plugin 0.24.0 | 0.34.0 |
| CI on JDK 11 | CI on JDK 17 |

Everything the upgrade forced:

- **`namespace`** — AGP 8 takes the R class package from the build script, not the manifest.
  `library/src/main/AndroidManifest.xml` held nothing but `package=`, so it is deleted; the
  sample keeps its manifest minus that attribute.
- **ViewBinding** — `kotlin-android-extensions` was removed in Kotlin 1.9, so
  `MainActivity` no longer uses synthetic view accessors.
- **ConstraintLayout** — the sample pulled it from the retired
  `com.android.support.constraint` coordinates and relied on Jetifier to rewrite it to
  androidx. It now depends on androidx directly, and `android.enableJetifier` is off.
  Pinned at 2.1.4 because 2.2.x requires minSdk 21 and this project's floor is 19.
- **jvmTarget** — Java source/target stays at 1.8, so both modules pin the Kotlin target to
  match. Without it AGP 8 fails the build on the mismatch.
- **Mockito 5** — 4.7 cannot instrument classes on JDK 17+; 42 of the 58 unit tests died
  with `MockitoException` before this. Mockito 5 makes the inline mock maker the default,
  which is what `mockito-inline` was there for.
- **ktlint 1.x rules** — a KDoc comment is no longer allowed at file top level, so the
  license header is a plain `/*` block; the seven wildcard imports are expanded; and
  `emptyChar` keeps its name under a targeted `@Suppress`, since renaming it to
  `EMPTY_CHAR` would break consumers of the published library.

The reformatting ktlint 1.8 asks for is a **separate second commit**, so the first commit
reads without 500 lines of trailing commas in the way.

## Fix for #8

`scripts/pre-commit` called `./gradlew ktlint` — a task this project has never registered,
since ktlint appears only as a step inside the Spotless configuration. Every commit made
with the hook installed was therefore rejected. It now calls `:library:spotlessCheck`.

## Verification

`./gradlew build` — the exact command CI runs — passes on JDK 21, including `spotlessCheck`
and all 58 unit tests. The instrumented tests under `src/androidTest` were **not** run: they
need a device, and none was available.

## Not in scope

Publishing is broken: `publish.yml` calls Gradle tasks that the new maven-publish plugin no
longer registers, and `SONATYPE_HOST=S01` points at a retired host. Filed as #9, to be fixed
in a follow-up.

https://claude.ai/code/session_01A1A3KFw9GQ8RGP4eeNnq6j

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -9,11 +9,10 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:7.4.2'
-        classpath "com.diffplug.spotless:spotless-plugin-gradle:4.0.0"
-        classpath 'org.jetbrains.kotlin:kotlin-gradle-plugin:1.7.21'
-        classpath 'com.vanniktech:gradle-maven-publish-plugin:0.24.0'
-        classpath 'com.android.tools.build:gradle:7.4.2'
+        classpath 'com.android.tools.build:gradle:8.13.0'
+        classpath "com.diffplug.spotless:spotless-plugin-gradle:8.9.0"
+        classpath 'org.jetbrains.kotlin:kotlin-gradle-plugin:2.1.10'
+        classpath 'com.vanniktech:gradle-maven-publish-plugin:0.34.0'
     }
 }
 
--- a/scripts/pre-commit
+++ b/scripts/pre-commit
@@ -4,7 +4,6 @@ git stash -q --keep-index
 
 echo "Running git pre-commit hook"
 
-./gradlew :library:spotlessApply
 ./gradlew :library:spotlessCheck
 
 RESULT=$?
--- a/scripts/pre-commit
+++ b/scripts/pre-commit
@@ -5,7 +5,7 @@ git stash -q --keep-index
 echo "Running git pre-commit hook"
 
 ./gradlew :library:spotlessApply
-./gradlew ktlint
+./gradlew :library:spotlessCheck
 
 RESULT=$?
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,12 +8,13 @@ jobs:
 
     steps:
       - name: Checkout code
-        uses: actions/checkout@v2
+        uses: actions/checkout@v4
 
-      - name: Set up JDK 11
-        uses: actions/setup-java@v1
+      - name: Set up JDK 17
+        uses: actions/setup-java@v4
         with:
-          java-version: 1.11
+          distribution: 'temurin'
+          java-version: '17'
 
       - name: Build and Run Tests with Gradle
         run: ./gradlew build
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -13,13 +13,13 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
 
-      - name: Install JDK 11
-        uses: actions/setup-java@v3
+      - name: Install JDK 17
+        uses: actions/setup-java@v4
         with:
-          distribution: 'zulu'
-          java-version: 11
+          distribution: 'temurin'
+          java-version: '17'
 
       - name: Upload release
         run: ./gradlew publishAllPublicationsToMavenCentralRepository --no-daemon --no-parallel
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -88,6 +88,7 @@ A version lives in three files and all of them must be bumped together:
 2. `gradle.properties` — `VERSION_NAME`
 3. `history.txt` — one changelog line
 
-Publishing runs from `.github/workflows/publish.yml` on a GitHub release
-(`publishAllPublicationsToMavenCentralRepository`, then `closeAndReleaseRepository`); credentials
-come from repository secrets.
+Publishing runs from `.github/workflows/publish.yml` on a GitHub release; credentials come from
+repository secrets. **The workflow is currently broken**: it still calls the two Gradle tasks the
+maven-publish plugin dropped before 0.34.0, and `SONATYPE_HOST=S01` points at the retired OSSRH
+host. Migrating it to the Central Portal is tracked in issue #9.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,93 @@
+# CLAUDE.md
+
+Guidance for Claude Code when working in this repository.
+
+## What this is
+
+`ru.pravbeseda:CurrencyEditText` — an Android library (published to Maven Central) providing two
+views that format numeric/monetary input as the user types:
+
+- `CurrencyEditText` — extends `AppCompatEditText`.
+- `CurrencyMaterialEditText` — extends `TextInputLayout` and wraps a `CurrencyEditText` instance;
+  it delegates almost every method to that inner view and additionally renders validation errors
+  via `TextInputLayout.setError`.
+
+Gradle modules: `:library` (the published artifact) and `:sample` (demo app).
+
+## Build & test
+
+```bash
+./gradlew build                        # what CI runs (JDK 17)
+./gradlew :library:test                # JVM unit tests — the main suite
+./gradlew :library:connectedAndroidTest # instrumented tests, needs a device/emulator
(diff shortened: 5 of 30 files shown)


pravbeseda (commented):
🤖 Reviewed the toolchain upgrade (Gradle 8.13 / AGP 8.13 / Kotlin 2.1.10), the ktlint 1.8 reformat, the sample viewBinding migration and the pre-commit hook change: 2 blocking findings, 0 suggestions.

scripts/pre-commit
```diff
@@ -5,7 +5,7 @@ git stash -q --keep-index
 echo "Running git pre-commit hook"
 
 ./gradlew :library:spotlessApply
-./gradlew ktlint
+./gradlew :library:spotlessCheck
```
pravbeseda: 🤖 blocking: I am not sure this pair can ever reject a commit. `spotlessApply` on line 7 rewrites the working-tree files first, so by the time `spotlessCheck` runs there is nothing left for it to complain about — `RESULT` looks like it will be 0 for every autocorrectable violation, which is the whole class of problems the hook is there for.

And the files `spotlessApply` just rewrote are never added to the index, so `git commit` still records the unformatted staged content while the reformatting stays behind as an unstaged diff. That would mean staging a badly formatted `.kt` file, seeing the hook report success, committing it unformatted, and only finding out when `./gradlew build` fails `spotlessCheck` in CI — the exact outcome the hook was meant to prevent.

Is that the intended behaviour here, or was the hook supposed to block the commit / carry its fixes into it?

build.gradle
```diff
@@ -9,11 +9,10 @@ buildscript {
...
     dependencies {
-        classpath 'com.android.tools.build:gradle:7.4.2'
-        classpath "com.diffplug.spotless:spotless-plugin-gradle:4.0.0"
-        classpath 'org.jetbrains.kotlin:kotlin-gradle-plugin:1.7.21'
-        classpath 'com.vanniktech:gradle-maven-publish-plugin:0.24.0'
-        classpath 'com.android.tools.build:gradle:7.4.2'
+        classpath 'com.android.tools.build:gradle:8.13.0'
+        classpath "com.diffplug.spotless:spotless-plugin-gradle:8.9.0"
+        classpath 'org.jetbrains.kotlin:kotlin-gradle-plugin:2.1.10'
+        classpath 'com.vanniktech:gradle-maven-publish-plugin:0.34.0'
```
pravbeseda: 🤖 blocking: With the plugin at 0.34.0 I cannot find the two tasks the release workflow runs any more. On this commit:

```
$ ./gradlew :library:publishAllPublicationsToMavenCentralRepository --dry-run
* What went wrong:
Cannot locate tasks that match ':library:publishAllPublicationsToMavenCentralRepository' as task
'publishAllPublicationsToMavenCentralRepository' not found in project ':library'.
```

`./gradlew tasks --all` lists only `publish`, `publishMavenPublicationToMavenLocal` and `publishToMavenLocal` — no Maven Central target at all, and no `closeAndReleaseRepository` either. Both steps of `.github/workflows/publish.yml` (lines 25 and 33) name tasks that no longer exist, so the next GitHub release looks like it would fail on the first `./gradlew` call.

CI does not cover this — it only runs `./gradlew build` — so nothing on this PR would have caught it. The `SONATYPE_HOST=S01` in `gradle.properties` may be related, since the newer plugin versions moved to the Central Portal. Was the release path re-checked after the version bump? The new `CLAUDE.md` also documents the old two-task flow under "Releasing", so it would go stale with it.

scripts/pre-commit
pravbeseda: Confirmed, and fixed by dropping `spotlessApply` from the hook — line 7 is gone, so `spotlessCheck` is now the only Gradle call and its exit code is what `RESULT` captures. The hook rejects the commit and the developer runs `./gradlew :library:spotlessApply` themselves.

One correction to the finding: the hook was not completely toothless before. `spotlessApply` only fixes autocorrectable violations, so a non-autocorrectable ktlint rule still reached `spotlessCheck` and failed the commit. The part that held fully is the second one — the reformatted files were never staged, so the commit recorded the unformatted staged content while the fixes stayed behind as an unstaged diff.

Re-staging the fixes instead (`git add` after `spotlessApply`) was considered and rejected: it does not compose with the `git stash --keep-index` / `git stash pop` pair on lines 3 and 11, which would then have to pop over files the hook had just rewritten.

build.gradle
pravbeseda: Confirmed against the build at this head — `./gradlew :library:publishAllPublicationsToMavenCentralRepository --dry-run` fails with "task not found", and neither task is registered any more.

The release path itself is deferred, not dismissed: it is already tracked in #9, with the Central Portal migration spelled out there. Two reasons it does not belong in this PR: `SONATYPE_HOST=S01` points at the OSSRH host Sonatype retired in June 2025, so publishing is broken on `main` too and reverting the plugin bump would not restore it; and the actual fix needs Central Portal credentials in the repository secrets, so it could only land here unverified.

What was in scope and is fixed here: the `Releasing` section of the `CLAUDE.md` this PR adds no longer documents the two non-existent tasks — it now says the workflow is broken and points at #9.

PR #11, opened 2026-09-06
fix: publish releases through the Central Portal from a manual workflow

### LINKED ISSUE #9: Publishing workflow calls Gradle tasks that no longer exist after the maven-publish plugin upgrade

pravbeseda:
## Problem

`SONATYPE_HOST=S01` is no longer a value the maven-publish plugin recognises. Version 0.34.0 calls
`publishToMavenCentral()` only when the property equals `CENTRAL_PORTAL`, so with `S01` the plugin
configured no Maven Central publishing at all:

```
./gradlew :library:tasks --all | grep -i publish
publish
publishMavenPublicationToMavenLocal
publishToMavenLocal
```

That is why `publishAllPublicationsToMavenCentralRepository` was "not found" — the task was never
registered. `closeAndReleaseRepository` is gone for good; the plugin now releases from a single
task.

## Change

- `gradle.properties`: `SONATYPE_HOST=S01` → `CENTRAL_PORTAL`. All Central tasks appear again,
  including `publishAndReleaseToMavenCentral`.
- New `.github/workflows/release.yml`, started by hand from the Actions tab with a
  `patch` / `minor` / `major` choice. It bumps the version, builds and tests, publishes to Maven
  Central, then pushes the bump commit and the tag to `main` and creates a GitHub release with
  generated notes. `publish.yml` is removed.
- The workflow refuses to run from any branch but `main`, checks that `main` has not moved before
  the irreversible publish, pushes `--atomic`, and takes a `concurrency` group so two release runs
  cannot overlap.
- New `scripts/bump-version.sh`, which raises `VERSION_NAME` and `VERSION_CODE` in
  `gradle.properties` and reports the new version to the workflow.
- The version now lives in exactly one file. `dependencies.gradle` is deleted along with the four
  `apply from:` lines that pulled it in; `sample/build.gradle` reads `VERSION_NAME` and
  `VERSION_CODE` directly, and the unused `publishGroupID` / `publishArtifactID` keys go with it.
- `history.txt` is removed: the generated release notes replace it. Nothing in the build or in the
  POM ever referenced the file.
- `CLAUDE.md` Releasing section rewritten for the new process.

Publishing runs before anything is pushed, because a Central release cannot be undone: if it fails,
the repository is left untouched and the run can simply be repeated.

Tags keep the existing unprefixed form (`1.0.4`), matching every tag since 0.4.0.

## Verification

- `./gradlew build` — green.
- `./gradlew :library:tasks --all` now lists `publishToMavenCentral` and
  `publishAndReleaseToMavenCentral`.
- `./gradlew :library:publishToMavenLocal` produces aar, sources jar, javadoc jar and pom.
- `scripts/bump-version.sh` exercised for patch, minor and major on a copy of the repository
  (1.0.3 → 1.0.4 / 1.1.0 / 2.0.0, version code 19 → 20); `shellcheck` clean.
- The sample's merged manifest still carries `versionCode="19"` / `versionName="1.0.3"` after the
  move to `gradle.properties`.

## Follow-up outside this PR

`OSSRH_USERNAME` and `OSSRH_PASSWORD` must be re-issued as a Central Portal user token before the
next release; the old OSSRH credentials do not work against the Portal.

Closes #9


.github/workflows/release.yml
```diff
@@ -0,0 +1,68 @@
...
+
+jobs:
+  release:
+
+    runs-on: ubuntu-latest
+    if: github.repository == 'pravbeseda/CurrencyEditText'
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
```
pravbeseda: 🤖 blocking: Could a manual run from a feature branch publish and land that branch's unreleased changes directly on `main`? The [Run workflow UI allows choosing a branch](https://docs.github.com/en/actions/how-tos/manage-workflow-runs/manually-run-a-workflow), and checkout defaults to that event's ref. The job only checks the repository name, but later unconditionally pushes `HEAD:main`. For a feature branch ahead of the current `main`, a patch run therefore publishes its code to Central and fast-forwards `main` to include all its commits, without merging its PR (`main` is currently unprotected). This makes the selected release source and the branch being updated disagree.

.github/workflows/release.yml
```diff
@@ -0,0 +1,68 @@
...
+
+      - name: Commit and tag
+        env:
+          VERSION: ${{ steps.bump.outputs.version }}
+        run: |
+          git config user.name 'github-actions[bot]'
+          git config user.email 'github-actions[bot]@users.noreply.github.com'
+          git commit -m "chore: release $VERSION" gradle.properties dependencies.gradle
+          git tag "$VERSION"
+          git push origin HEAD:main "$VERSION"
```
pravbeseda: 🤖 blocking: What happens if an ordinary PR is merged into `main` while this run is building or publishing? Starting at version `1.0.3`, the run publishes `1.0.4`, then creates a bump commit on the old checkout; the new remote `main` is not its ancestor, so this push rejects the branch update. Because the push is not atomic, the new `1.0.4` tag can still be created, while `main` retains `1.0.3` and the GitHub release step is skipped. A fresh patch run then tries to publish the already released `1.0.4` again. I am not sure how this workflow can complete that release after this reachable failure: [Git rejects non-fast-forward branch updates](https://git-scm.com/docs/git-push#_note_about_fast_forwards), and [Central does not allow replacing a released version](https://central.sonatype.org/faq/can-i-change-a-component/).

pravbeseda (commented):
🤖 Reviewed the full PR at 94db96861e05665e80de6dc630705861e2f5bfc1; 2 blocking findings and 0 suggestions follow.

.github/workflows/release.yml
pravbeseda: 🤖 blocking: `workflow_dispatch` can be started from any ref — the branch picker in the Actions tab is the only thing that decides which — and nothing in this job ties the run to `main`. Dispatched from a feature branch that is ahead of `main`, this line fast-forwards `main` onto that branch's tip, so unreviewed commits land on the default branch; and by then the previous step has already released that same branch's code to Maven Central under the new version, which cannot be taken back. The old `publish.yml` could not do either, since it only ever reacted to a release and pushed nothing. Am I missing something that constrains the ref here, or is running from a non-`main` branch simply not expected to happen?

scripts/bump-version.sh
```diff
@@ -0,0 +1,27 @@
...
+case "$increment" in
+    major) major=$((major + 1)); minor=0; patch=0 ;;
+    minor) minor=$((minor + 1)); patch=0 ;;
+    patch) patch=$((patch + 1)) ;;
+    *) echo "unknown increment: $increment" >&2; exit 1 ;;
+esac
+version="$major.$minor.$patch"
+
+sed -i.bak "s/^VERSION_NAME=.*/VERSION_NAME=$version/" gradle.properties
+sed -i.bak -e "s/^\( *publishVersion *: *\)'.*'/\1'$version'/" \
```
pravbeseda: 🤖 suggestion: the version number now exists in two places and this script's job is partly to keep them equal — `VERSION_NAME` in `gradle.properties`, which the publish plugin reads, and `publishVersion` in `dependencies.gradle`, whose only consumer is `sample/build.gradle:24`. Since properties from `gradle.properties` are visible to every build script anyway, I am not sure the second copy is earning its keep: dropping it would take this `-e` expression and one line of `dependencies.gradle` with it, and leave `publishVersionCode` as the only thing that file has to carry. Not a blocker either way — your call.

pravbeseda (commented):
🤖 Reviewed the release migration at 94db968 (release workflow, bump script, SONATYPE_HOST, docs): 1 blocking finding and 1 suggestion follow.

.github/workflows/release.yml
pravbeseda: Fixed. The job now opens with a guard step that fails the run when `github.ref` is not
`refs/heads/main`, so a dispatch from any other branch stops before the build, the publish and the
push.

Chosen over a job-level `if:` condition, because a skipped job reads like a success in the Actions
UI and says nothing about why nothing happened; and over pinning the checkout to `ref: main`,
because that would silently release code other than the branch the operator picked in the dropdown.

Note: two independent review runs raised this same finding on two threads. Both are answered by
this one change.

.github/workflows/release.yml
pravbeseda: Fixed. The job now opens with a guard step that fails the run when `github.ref` is not
`refs/heads/main`, so a dispatch from any other branch stops before the build, the publish and the
push.

Chosen over a job-level `if:` condition, because a skipped job reads like a success in the Actions
UI and says nothing about why nothing happened; and over pinning the checkout to `ref: main`,
because that would silently release code other than the branch the operator picked in the dropdown.

Note: two independent review runs raised this same finding on two threads. Both are answered by
this one change.

pravbeseda: Confirmed and fixed, keeping the publish-before-push order.

Three changes:

- A `Check that main has not moved` step now runs immediately before the publish: it fetches
  `origin/main` and fails the run unless it still equals `$GITHUB_SHA`. The window in which a
  merge can invalidate the push shrinks from the whole build-and-publish, minutes, to the seconds
  between that check and the push.
- The push is now `--atomic`, so the partial outcome you describe — the tag created while the
  branch update is rejected — can no longer happen; either both refs land or neither does.
- A workflow-level `concurrency: group: release` keeps two release runs from overlapping each
  other.

The order itself stays as it is on purpose. Reversing it would remove the race completely, but it
would trade it for the more likely failure: when publishing fails, a tag and a GitHub release would
already exist for a version that never reached Central. That is exactly the state this repository
is in for 1.0.3, which the previous, broken workflow left behind.

scripts/bump-version.sh
pravbeseda: Agreed, and taken further than proposed. Dropping only `publishVersion` would have left the version
split across two files by field — the name in `gradle.properties`, the code in `dependencies.gradle`
— which is a stranger state than either end of the choice.

So the whole duplication is gone:

- `VERSION_CODE` moved into `gradle.properties` next to `VERSION_NAME`; that file is now the single
  place a version lives.
- `sample/build.gradle` reads `VERSION_NAME` and `VERSION_CODE.toInteger()` directly.
- `dependencies.gradle` is deleted, together with the four `apply from:` lines that pulled it into
  the root, library and sample scripts. Nothing else in it was still used.
- `scripts/bump-version.sh` now edits one file instead of two, and the workflow commits one path.

Verified: `./gradlew build` is green, and the sample's merged manifest still carries
`versionCode="19"` / `versionName="1.0.3"`.

.github/workflows/release.yml
pravbeseda: 🤖 reopened at e68fe3c6afed0ecb97d42af043ddea26244c7ed7

The new pre-publish check covers a merge during the build, and `--atomic` covers the partial branch/tag update. However, the original finding also named a merge during publishing, and that case still appears reachable: the check at lines 60–64 passes at commit A with version 1.0.3; an ordinary PR then advances `main` to B while line 67 publishes 1.0.4; the release bump commit is still based on A, so the push at line 82 is rejected. Atomicity now leaves neither the bump nor the tag on GitHub, but 1.0.4 remains published on Central, `main` still declares 1.0.3, and the next patch run selects the already published 1.0.4 again.

The interval includes the entire publishing step, not just the time after publishing. The release concurrency group serializes release jobs; it does not prevent an ordinary PR merge. This is the still-uncovered case from the original finding, rather than an objection to the chosen publish-before-push order. Am I missing anything that lets this release complete when `main` advances after the check succeeds?

References: [Git push rejection and atomic updates](https://git-scm.com/docs/git-push), [workflow concurrency scope](https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-workflow-concurrency), and [Central release immutability](https://central.sonatype.org/faq/can-i-change-a-component/).

pravbeseda: You are right, and my previous reply was wrong about the size of the window. The check sits before
the publish, so the interval it leaves open is check → push, which contains the whole publishing
step — not the seconds I claimed.

Fixed by making the rejected push impossible rather than rare. The `Commit and tag` step now
rebases before pushing:

```
git tag "$VERSION"
git fetch origin main
git rebase origin/main
git push --atomic origin HEAD:main "$VERSION"
```

`fetch-depth: 0` is back on the checkout, because a rebase needs the history a shallow clone does
not have.

In the scenario you describe — the check passes at A, an ordinary PR advances `main` to B while
1.0.4 is publishing — the bump commit is now replayed onto B and the push is a fast-forward, so the
release completes: `main` declares 1.0.4 and the GitHub release step runs. The tag is created
before the rebase, so it stays on the commit the published artifact was actually built from, rather
than on a tree carrying B's changes that never reached Central.

What is left is the seconds between the rebase and the push, which is the window I wrongly claimed
last time.

The pre-publish check stays, for the other reason it turned out to be worth having: if `main` moved
during the build, the run would publish an artifact that already omits what was merged. Failing
there is better than releasing it.

PR #19, opened 2026-09-06
chore: set up quality gates — version catalog, detekt, Kover, Lint, CI

### LINKED ISSUE #15: Set up quality gates: version catalog, detekt, Kover, Lint baselines, CI
## Why

Code quality here rests on one gate (Spotless) and on manual review. Three things are missing and each one is cheap to add:

- **Coverage is not measured at all.** There is no number, so there is nothing to notice when it drops.
- **Complexity is not bounded.** `CurrencyInputWatcher.onTextModified()` (`library/src/main/java/ru/pravbeseda/currencyedittext/watchers/CurrencyInputWatcher.kt:51`) is 145 lines with ~25 branches, and it is the function nearly every change in this repo touches.
- **Lint findings block nothing.** Lint already runs inside `check` (verified with `--dry-run`), but its 11 warnings in `:library` and 18 in `:sample` are advisory, so they accumulate unread.

The stack below is the one already running in the author's other Android project, so it is a port rather than an experiment.

## Scope

One PR. It touches no production Kotlin code.

### Build infrastructure

- [ ] `gradle/libs.versions.toml` — every version from the three `build.gradle` files moves here; root script switches from `buildscript { classpath … }` to the `plugins { alias(…) }` DSL.
- [ ] Git hooks move to a versioned `.githooks/` directory, with `core.hooksPath` pointed at it from `settings.gradle` at settings evaluation (skipped when `CI` is set). Replaces the `installGitHook` copy task, which only works if someone remembers to run it.

### Gates

- [ ] **detekt 1.23.8** (last stable line; 2.x has been in alpha since 2025 and needs Gradle 8.14.3+, this project is on 8.13). `config/detekt/detekt.yml` on top of `buildUponDefaultConfig`; per-variant baseline freezing what exists today; run with type resolution and wire those tasks into `check`, since the plain `detekt` task analyses far less. Applied to `:library` and `:sample`.
- [ ] **Kover 0.9.9** on `:library` only, and filtered by class:

```groovy
kover {
    reports {
        filters { includes { classes('ru.pravbeseda.currencyedittext.watchers.*', 'ru.pravbeseda.currencyedittext.util.*') } }
        variant('debug') {
            log { onCheck = true }
            verify { onCheck = true; rule { minBound(N) } }
        }
    }
}
```

  The filter is the point: `CurrencyEditText` and `CurrencyMaterialEditText` cannot be reached from `src/test`, so an unfiltered figure would measure the share of View code in the module rather than the quality of the tests. Set `N` from the first measured run, minus a small margin.
- [ ] **Android Lint**: `lint { abortOnError true; warningsAsErrors true }` plus a `lint-baseline.xml` per module holding today's findings.
- [ ] **Spotless**: unchanged behaviour, ktlint version read from the catalog.

### CI

- [ ] `pull_request` trigger in addition to `push` (today it is `on: [push]` only).
- [ ] `concurrency` group keyed by ref, cancel-in-progress for pull requests only.
- [ ] `gradle/actions/setup-gradle` for the dependency cache.
- [ ] Steps: `spotlessCheck`, unit tests, `koverVerify` + `koverLog`, detekt, lint — each blocking, test and lint reports uploaded as artifacts.

### Rules for the agent (CLAUDE.md)

- [ ] **TDD as a rule, not a preference**, with a Definition of Done: the local command that must pass; a test written and seen failing before the implementation; a bug fix starting from a test that reproduces the bug and names the issue; behaviour changes reflected in tests; logic that cannot be tested from `src/test` being a signal to move it out of the View, not a licence to skip the test.
- [ ] **Never weaken a check to make it pass** — no `@Suppress`, no new baseline entries, no lowered coverage bound, no `@Ignore` without explicit permission. Baselines only ever shrink.
- [ ] Do not grow `onTextModified`: a new formatting rule is a separate named step, not another `if` in the middle.
- [ ] No `Array<Any?>` and no casting results back out — return a data class.
- [ ] No silently swallowed exceptions: a `catch` that returns a default needs a comment saying why losing the information is safe.
- [ ] Public API changes in `ru.pravbeseda.currencyedittext` get their own "Public API change" bullet in the PR description — the artifact goes to Maven Central and a release cannot be taken back.
- [ ] Test libraries belong in `testImplementation`/`androidTestImplementation`, never in `implementation` (this repo has already shipped that bug — 6b75231).
- [ ] Add a warning to `.githooks/pre-commit` when a commit touches `library/src/main/**` without touching `library/src/test/**`.

## Done when

`./gradlew build` and CI are green with all gates blocking, every existing finding frozen in a baseline, the coverage figure printed in the build log, and a fresh clone getting the git hooks without anyone running a task by hand.

pravbeseda:
Closes #15.

Code quality here rested on one gate (Spotless) and on manual review. This adds four blocking gates and the infrastructure they need. **No production Kotlin is touched** — the only files under `src/` that change are none; everything is build scripts, configuration, baselines and docs.

## What now blocks a build

`./gradlew build` runs all five:

| Gate | Configured in | Frozen |
|---|---|---|
| Spotless (ktlint, version from the catalog) | `spotless.gradle` | — formatting is fixed, not frozen |
| detekt 1.23.8, type resolution, both modules | `config/detekt/detekt.yml`, root `build.gradle` | 35 findings across 4 baselines |
| Kover 0.9.9, filtered by class | `library/build.gradle` | 86.08% today, `minBound(80)` |
| Android Lint, `warningsAsErrors` | root `build.gradle` | 2 + 28 findings |
| Unit tests | `library/src/test` | — |

Each gate was checked against a probe rather than assumed:

- a temporary `TODO:` in `Routines.kt` → `detektDebug` red on `ForbiddenComment`;
- a temporary `android:text="…"` in the sample layout → `lintDebug` red on `HardcodedText`;
- staging a change under `library/src/main` alone → the hook prints the missing-test warning, and `spotlessCheck` blocks on the formatting violation in the same run.

## Build infrastructure

- **`gradle/libs.versions.toml`** now holds every version. The root script moves from `buildscript { classpath … }` to the `plugins { alias(…) }` DSL, which needs the `pluginManagement` block added to `settings.gradle`: AGP resolves from Google's Maven, which the plugin portal does not mirror.
- **Git hooks move to `.githooks/`**, versioned, with `core.hooksPath` pointed at them from `settings.gradle` at settings evaluation (skipped when `CI` is set). This replaces the `installGitHook` copy task, which only worked if someone remembered to run it — a fresh clone now gets the hook without anyone running a task by hand.

## Three decisions worth a look

**1. Coverage bound is 80 against a measured 86.08%.** The filter is the point: `CurrencyEditText` and `CurrencyMaterialEditText` cannot be reached from `src/test`, so an unfiltered figure would measure the share of View code in the module rather than the quality of the tests, and would sink whenever a View grew. The bound covers `watchers.*` and `util.*` only. Six points of margin so a branch landing with its edge cases covered a commit later does not turn the floor into an argument with the build; it rises when the margin turns out to be unused.

**2. `GradleDependency` and `NewerVersionAvailable` are demoted to `informational`, `AndroidGradlePluginVersion` is disabled.** These fire when a newer version exists in the world, not when anything in the diff is wrong, and their messages carry the newer version number — so every frozen baseline entry stops matching the day an upstream release lands, and a branch that changed nothing goes red for reasons no diff explains. Demoted rather than disabled, so an ageing dependency still shows up in the lint report CI uploads; it just cannot fail a build on the calendar's say-so. Dependency updates belong to Renovate, which is a later stage.

**3. The pre-commit hook no longer stashes.** The old `scripts/pre-commit` ran `git stash --keep-index` and popped afterwards; a hook that dies between the stash and the pop is a known way to lose work. `spotlessCheck` now reads the working tree instead. The hook also warns — does not block — when a commit touches `library/src/main` without `library/src/test`, so skipping a test is a visible act rather than an unnoticed one.

## The plain `detekt` task is disabled

It analyses on the PSI alone, and the rules that matter most here (`UnsafeCallOnNullableType`, `UseOrEmpty` — 13 of the 25 frozen findings in `:library`) need type resolution to fire at all. Leaving it enabled would offer a green run that checks a fraction of what the gate checks. `check` and `detektAll` are both wired from one list of typed tasks, so the two cannot drift.

## CI

`pull_request` trigger alongside `push`; a `concurrency` group keyed by ref that cancels in-progress runs for pull requests only; `gradle/actions/setup-gradle` for the dependency cache; one blocking step per gate, each with `if: always()` so a red gate never hides the verdict of the others; test, detekt and lint reports uploaded as artifacts.

## CLAUDE.md

A **Quality gates** section, a **Definition of Done** that makes test-first a rule rather than a preference, and the rules the gates cannot express: never weaken a check to make it pass (baselines only shrink), do not grow `onTextModified`, no `Array<Any?>`, no silently swallowed exceptions, public API changes called out in the PR description, test libraries never in `implementation`.

## Public API change

None. No file under `library/src/main/java` is modified.

## What is deliberately not here

Stages 4 and 6–8 of the roadmap in #15's discussion: running the instrumented tests on an emulator in CI, Renovate, the refactoring of `onTextModified` under the now-green gates, and `dependency-analysis` / binary-compatibility validation. Each is its own change.


build.gradle
```diff
@@ -1,36 +1,135 @@
...
+// checking different things. The release variant is left out: neither module has
+// release-specific Kotlin, so it shares its sources with debug.
+def typedDetektTasks = [
+    ':library:detektDebug',
+    ':library:detektDebugUnitTest',
+    ':library:detektDebugAndroidTest',
+    ':sample:detektDebug',
+    ':sample:detektDebugUnitTest',
+    ':sample:detektDebugAndroidTest',
+]
```
pravbeseda: 🤖 suggestion: this list names the two modules by hand, while `subprojects` a few lines below applies detekt to every module there will ever be. If a third module is added, `detektAll` skips it, and `typedDetektTasks.findAll { it.startsWith(":${project.name}:") }` on line 121 returns an empty list — so that module's `check` depends on no detekt task and passes having analysed nothing, which is the exact failure the comment on line 116 was written to prevent, only silent this time. Is there a reason to write the task paths out rather than derive them per project from the variants detekt has already created?

config/detekt/detekt.yml
```diff
@@ -0,0 +1,26 @@
...
+        value: 'TODO:'
+      - reason: 'Reference an issue instead, so the debt is tracked outside the code.'
+        value: 'FIXME:'
+  MagicNumber:
+    # Off in tests: an expected value written as a literal is the point of the test.
+    excludes: ['**/test/**', '**/androidTest/**']
+
+exceptions:
+  TooGenericExceptionCaught:
+    excludes: ['**/test/**', '**/androidTest/**']
```
pravbeseda: 🤖 suggestion: under `buildUponDefaultConfig` detekt replaces a list-valued key rather than merging into it, so these three blocks narrow the defaults instead of adding to them. `ForbiddenComment.comments` drops `STOPSHIP:`, which 1.23.8 forbids out of the box. `MagicNumber.excludes` and `TooGenericExceptionCaught.excludes` both already default to `['**/test/**', '**/androidTest/**', '**/commonTest/**', '**/jvmTest/**', '**/androidUnitTest/**', '**/androidInstrumentedTest/**', '**/jsTest/**', '**/iosTest/**']` (plus `**/*.kts` for `MagicNumber`), so restating the first two entries only takes the remaining ones away — the two `excludes` blocks look like they could go entirely and leave the behaviour they document intact. Was the narrowing intended, or did these mean to restate a default rather than override it?

sample/lint-baseline.xml
```diff
@@ -0,0 +1,279 @@
...
+        errorLine1="        targetSdkVersion 34"
+        errorLine2="        ~~~~~~~~~~~~~~~~~~~">
+        <location
+            file="build.gradle"
+            line="22"
+            column="9"/>
+    </issue>
+
+    <issue
+        id="GradleDependency"
```
pravbeseda: 🤖 suggestion: these 13 entries — and `library/lint-baseline.xml:16` — freeze `GradleDependency` and `NewerVersionAvailable`, the two ids `build.gradle:89` demotes to `informational` specifically so that they "stay in the lint report CI uploads, so an ageing dependency is still visible". As I read it a baselined finding is filtered out of the report rather than downgraded in it, so the uploaded report currently shows none of them, and each entry carries the version number the same comment says will stop matching the day an upstream release lands. Would dropping them leave the demotion doing what the comment describes?

CLAUDE.md
```diff
@@ -17,18 +17,48 @@ Gradle modules: `:library` (the published artifact) and `:sample` (demo app).
...
+## Quality gates
+
+All five block the build, and `./gradlew build` runs all five:
+
+| Gate | Where it is configured | What is frozen |
+|---|---|---|
+| Spotless (ktlint + license header) | `spotless.gradle` | nothing — formatting is fixed, not frozen |
+| detekt 1.23.8, type resolution | `config/detekt/detekt.yml`, root `build.gradle` | `*/detekt-baseline-*.xml` |
+| Kover 0.9.9, `minBound(80)` | `library/build.gradle` | — the bound is a floor, not a baseline |
+| Android Lint, `warningsAsErrors` | `library/build.gradle`, `sample/build.gradle` | `*/lint-baseline.xml` |
```
pravbeseda: 🤖 suggestion: this row sends the reader to `library/build.gradle` and `sample/build.gradle`, but neither file mentions lint — the whole block lives in the root `build.gradle`, inside `plugins.withId('com.android.base')` (lines 63-91), which is also what the PR description says. Since this table is what the next reader searches before touching a gate, should the row point at the root script the way the detekt row above it does?

pravbeseda (commented):
🤖 Reviewed the quality-gate setup (build scripts, detekt/Lint/Kover config, baselines, CI, docs) — 0 blocking findings, 4 suggestions.

build.gradle
pravbeseda: 🤖 The silent-skip is real and is now fixed, but not by deriving the list.

Deriving from the tasks detekt registers was weighed and turned down on the ground that the written approach is equally valid: `tasks.withType(Detekt).matching { it.name.startsWith('detektDebug') }` covers a new Android module, but a module without the Android plugin registers `detektMain` / `detektTest` and would be skipped just as silently — the same defect in a different shape. Covering both needs the matcher to branch on whether the Android plugin is applied and to exclude `detektMain` on Android modules so the analysis is not run twice, which trades a list you can read at a glance for a condition you have to reason about.

What the comment actually names — that the skip is silent — is removed instead. The root script now fails configuration when a subproject contributes no task to the gate:

```
:sample contributes no task to the detekt gate. Add its typed detekt tasks to
typedDetektTasks in the root build script.
```

Verified by deleting the `:sample` entries and running `./gradlew help`, which produced exactly that message.

config/detekt/detekt.yml
pravbeseda: 🤖 Correct on both counts, and fixed. Checked against `default-detekt-config.yml` at v1.23.8:

- `ForbiddenComment.comments` defaults to `FIXME:`, `STOPSHIP:`, `TODO:`. The override kept two and so silently stopped forbidding `STOPSHIP:`. `STOPSHIP:` is restored with the same project-specific reason, and a comment now says why all three values are restated rather than only the ones that deviate.
- Both `excludes` blocks are gone. `MagicNumber` and `TooGenericExceptionCaught` default to eight test-path globs (plus `**/*.kts` for `MagicNumber`), which already covers what the two entries were narrowing them to. `./gradlew detektAll` stays green without them, which is the check that the behaviour they documented survives their removal.

sample/lint-baseline.xml
pravbeseda: 🤖 Correct, and fixed. The claim was checked against the report the CI step uploads rather than reasoned about: with the entries in place `lint-results-debug.xml` contained a single `LintBaseline` note and no findings at all, in both modules — so the demotion to `informational` was doing nothing the comment claimed for it.

The 14 `GradleDependency` / `NewerVersionAvailable` entries are dropped from both baselines. The same report now holds 9 + 4 of them in `:sample` and 1 in `:library`, as hints, and `./gradlew clean build` stays green — `warningsAsErrors` promotes warnings, not informational findings, so they are visible without being blocking. That is what the comment on the `informational` line describes.

CLAUDE.md
pravbeseda: 🤖 Correct, and fixed: the row now points at the root `build.gradle`, the way the detekt row above it does. The lint block was moved into the root script late, to stop the same 25 lines being maintained twice, and this table row was not carried along with it.

Issue #18, opened 2026-09-06
Deferred quality work: instrumented tests in CI, dependency analysis, ABI validation


pravbeseda:
## Why

Three improvements that were considered and deliberately deferred. Kept in one issue so they are not rediscovered from scratch in six months.

## 1. Run the instrumented tests in CI

`library/src/androidTest` holds 9 tests over the real `CurrencyEditText` and `CurrencyMaterialEditText`. CI only compiles them (`:library:compileDebugAndroidTestSources`), so they have never run, and the two View classes — about 525 lines of public API — are checked by nothing.

Closing this means an emulator job (`reactivecircus/android-emulator-runner`, one API level). Deferred because it adds roughly ten minutes to every CI run for code that changes rarely. Reconsider when a bug lands in either View.

## 2. `dependency-analysis` as a periodic audit

`com.autonomousapps.dependency-analysis` reports misdeclared dependencies — unused, used but undeclared, or on the wrong configuration. Relevant here: this repo has already shipped `androidx.test:monitor` on `implementation` of a published library (6b75231).

Wire it in as `./gradlew buildHealth`, run by hand, **not** as part of `check`: it has no baseline, and "unused" for a dependency pulled in for its resources is a false positive by construction.

## 3. Binary compatibility validation

The library publishes to Maven Central, where a release cannot be withdrawn, and nothing today catches an accidental signature change. Two options:

- `Kotlin/binary-compatibility-validator` — works on the current Kotlin 2.1.10, available now.
- The Kotlin Gradle Plugin's built-in `abiValidation {}` with `checkKotlinAbi` — the direction upstream is going, but the DSL is experimental and needs Kotlin 2.2+, so it waits for a toolchain upgrade.

Either way the API dump lands in the repository and a change to it becomes a reviewable diff.

PR #20, opened 2026-09-06
chore: freeze the published API with binary-compatibility-validator

pravbeseda:
## Why

The library publishes to Maven Central, where a release cannot be withdrawn, and nothing today
catches an accidental signature change. Every existing gate — Spotless, detekt, Kover, Lint, unit
tests — looks inside the code; none looks at its outward boundary, so a renamed public method or a
changed return type reaches users as a build that stops compiling.

Part 3 of #18. Parts 1 (instrumented tests on an emulator in CI) and 2 (`dependency-analysis` as a
periodic `buildHealth` audit) are untouched and the issue stays open.

## What changed

- `binary-compatibility-validator` 0.18.2 applied to `:library` only — `:sample` publishes nothing.
- `library/api/library.api` — the dump of the public API, 157 lines: both Views,
  `CurrencyFormatConfig`, `CurrencyInputWatcherConfig`, `Routines`, `CurrencyInputWatcher`,
  `EasyTextWatcher`.
- A `Check binary compatibility` step in `ci.yml`. `apiCheck` is wired into `check`, so
  `./gradlew build` and the release workflow already run it; the CI job runs the gate tasks one by
  one rather than `check`, so it needs the step spelled out.
- `CLAUDE.md`: the gates table goes from five to six, the command list gains `apiCheck` / `apiDump`,
  and the "Public API change" rule now also asks for the `apiDump` diff in the same commit.

## How an intended API change works from now on

`./gradlew :library:apiDump`, then commit the rewritten `library/api/library.api`. It is not a
baseline of accepted debt: the file is meant to change, and the point is that the change shows up
in the diff as a reviewable line instead of being noticed after publication.

## Plugin choice

The standalone plugin rather than the Kotlin Gradle Plugin's own `abiValidation {}`. The built-in
one is the direction upstream is going, but its DSL is experimental and needs Kotlin 2.2+, while
this project is on 2.1.10. The reason is recorded next to the version in `libs.versions.toml`, so
the decision is not rediscovered when the toolchain moves.

## Public API change

None. The dump was generated from the current sources and no signature was touched.

## Testing

- `./gradlew build` — green.
- The gate was watched failing first: a temporary `fun temporaryAbiProbe(): String` added to
  `util/Routines.kt` made `:library:apiCheck` fail and print the diff of the added line; the probe
  was then removed.
- No unit tests, because the change is entirely build scripts, CI configuration and a generated
  dump — there is no `library/src/main` logic to cover.

### DIFF
--- a/library/api/library.api
+++ b/library/api/library.api
@@ -1,0 +1,157 @@
+public class ru/pravbeseda/currencyedittext/CurrencyEditText : androidx/appcompat/widget/AppCompatEditText {
+	public static final field Companion Lru/pravbeseda/currencyedittext/CurrencyEditText$Companion;
+	public fun <init> (Landroid/content/Context;Landroid/util/AttributeSet;)V
+	public final fun getDecimalSeparator ()C
+	public final fun getDecimalZerosPadding ()Z
+	public final fun getEmptyStringForZero ()Z
+	public final fun getGroupingSeparator ()C
+	public final fun getNegativeValueAllow ()Z
+	public final fun getValue ()Ljava/math/BigDecimal;
+	public final fun getValue0 ()Ljava/math/BigDecimal;
+	public final fun isValid ()Z
+	public final fun isValidState ()Z
+	protected fun onFocusChanged (ZILandroid/graphics/Rect;)V
+	protected fun onSelectionChanged (II)V
+	public final fun onValueChanged (Lkotlin/jvm/functions/Function3;)V
+	public final fun setCurrencySymbol (Ljava/lang/String;Z)V
+	public static synthetic fun setCurrencySymbol$default (Lru/pravbeseda/currencyedittext/CurrencyEditText;Ljava/lang/String;ZILjava/lang/Object;)V
+	public final fun setDecimalZerosPadding (Z)V
+	public final fun setEmptyStringForZero (Z)V
+	public final fun setLocale (Ljava/lang/String;)V
+	public final fun setLocale (Ljava/util/Locale;)V
+	public final fun setMaxNumberOfDecimalPlaces (I)V
+	public final fun setNegativeValueAllow (Z)V
+	public final fun setSeparators (CC)V
+	public fun setText (Ljava/lang/CharSequence;Landroid/widget/TextView$BufferType;)V
+	public final fun setValidator (Lkotlin/jvm/functions/Function1;)V
+	public final fun setValue (Ljava/math/BigDecimal;)V
+	public final fun setValue0 (Ljava/math/BigDecimal;)V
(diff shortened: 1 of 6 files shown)


library/api/library.api
```diff
@@ -0,0 +1,157 @@
...
+	public final fun getLocale ()Ljava/util/Locale;
+	public final fun getMaxNumberOfDecimalPlaces ()I
+	public final fun getNegativeValueAllow ()Z
+	public final fun getOnValueChanged ()Lkotlin/jvm/functions/Function1;
+	public fun hashCode ()I
+	public fun toString ()Ljava/lang/String;
+}
+
+public final class ru/pravbeseda/currencyedittext/util/CurrencyRounderKt {
+	public static final fun truncateNumberToMaxDecimalDigits (Ljava/lang/String;IC)Ljava/lang/String;
```
pravbeseda: 🤖 suggestion: this line and `firstChar` on line 138 look like helpers that leaked out rather than API anyone was meant to call: `truncateNumberToMaxDecimalDigits` has no caller outside `library/src/test`, and `firstChar` is used only from the two Views' `init` blocks. Now that the dump exists, `apiCheck` will defend both of them, and the new CLAUDE.md rule turns narrowing either one into a "Public API change" bullet — so whatever is in this file on merge is what the project has committed to keeping.

Were these two meant to be part of the published surface? If they were not, I would rather find that out now, while the file is being written for the first time, than the next time somebody wants to tidy `util/` up. (`Routines.bigDecimalToString` and `emptyChar` are clearly deliberate — the comment in `Routines.kt` says as much — so this is only about the other two.)

pravbeseda (commented):
🤖 Reviewed the whole PR — the plugin wiring, the dump, the CI step and the CLAUDE.md edits. No blocking findings; 1 suggestion follows.

library/api/library.api
pravbeseda: Correct on the facts — checked both: `truncateNumberToMaxDecimalDigits` has no caller in `library/src/main` at all (its only call site is `CurrencyRounderTest.kt:95`), and `firstChar` is called only from the two Views' `init` blocks. Neither carries the "Public API since 1.0" note that `emptyChar` has three lines above it, so neither looks deliberate.

Not acted on here, on scope: both are already public in the versions on Maven Central. This PR does not make them public — it records that they are. Narrowing either one to `internal` is itself a breaking change for anyone who happened to call it, so it needs a version decision and a release-notes line, and mixing "freeze the API as it is" with "change the API" into one PR makes both harder to verify: the first is checked by the dump matching the current code, which is exactly what would stop being true.

Filed as #22 with both file:line references and the release consideration, so the question survives past this thread. Resolving here.

Issue #23, opened 2026-09-06
parseMoneyValueWithLocale swallows ParseException and returns 0


pravbeseda:
## Why

`library/src/main/java/ru/pravbeseda/currencyedittext/util/Utils.kt:49` catches `ParseException`
and returns `0` with no comment. CLAUDE.md names this exact function as the example of how not to
handle a failure: "A `catch` that returns a default needs a comment saying why losing the
information is safe (`util/Utils.kt` `parseMoneyValue` is the example of how not to)."

The consequence is not local. `CurrencyEditText.stringToBigDecimal` (`CurrencyEditText.kt:243`)
feeds the result into `BigDecimal`, so `getValue()` — and therefore the `validator` a consumer
registers — cannot tell "the field holds nothing parseable" from "the field holds zero".

Split out of #17, which listed it and deliberately left it out: changing it changes behaviour for
consumers, so it needs its own decision and its own release call.

## Reproduction

```kotlin
val view = CurrencyEditText(context)
view.setText("")        // or any content the locale's NumberFormat cannot parse
view.getValue()         // BigDecimal.ZERO — indistinguishable from a field holding "0"
```

- Swallowed exception: `library/src/main/java/ru/pravbeseda/currencyedittext/util/Utils.kt:49`
- Where it reaches the public API: `library/src/main/java/ru/pravbeseda/currencyedittext/CurrencyEditText.kt:243`

## The decision this needs

1. Keep `0` and document why losing the exception is safe — a one-line comment at the `catch`, and
   `getValue()`'s contract stated as "an unparseable field reads as zero". Costs nothing, breaks
   nobody, satisfies the CLAUDE.md rule.
2. Return `null` from an unparseable field and let `getValue()` say so. Honest, but it changes the
   signature of a published API, so it belongs in the same major release as any other break.
3. Keep `0` but expose the distinction some other way (an `isValid`/`state` reading), leaving
   `getValue()` alone.

Whichever is chosen, the empty-field case deserves a test: today nothing pins down what an empty
`CurrencyEditText` reports.

## Done when

The `catch` either carries the reason it is safe or no longer swallows the information, the choice
is covered by a unit test, and `./gradlew build` is green.

Issue #22, opened 2026-09-06
Review the public surface of util/: truncateNumberToMaxDecimalDigits and firstChar


pravbeseda:
## Why

Raised while reviewing #20, and deliberately kept out of it: that PR records the public API as it
stands, and narrowing it is a different decision that needs its own release call.

Two declarations in `util/` are public without ever having been meant as API:

- `truncateNumberToMaxDecimalDigits` (`library/src/main/java/ru/pravbeseda/currencyedittext/util/CurrencyRounder.kt:31`)
  has no caller anywhere in `library/src/main`. Its only call site in the whole repository is
  `library/src/test/java/ru/pravbeseda/currencyedittext/CurrencyRounderTest.kt:95`.
- `firstChar` (`library/src/main/java/ru/pravbeseda/currencyedittext/util/Routines.kt:57`) is called
  only from the `init` blocks of the two Views — `CurrencyEditText.kt:87,89` and
  `CurrencyMaterialEditText.kt:77,79`.

Both now appear in `library/api/library.api` (lines 124 and 138), because that file records what is
public today rather than what was intended to be. `emptyChar`, three lines above `firstChar` in the
same file, carries the comment "Public API since 1.0: renaming it to EMPTY_CHAR would break
consumers" — so the deliberate part of that file was marked as such, and these two were not.

## The catch

They are already public in the versions on Maven Central. Narrowing either one to `internal` is
itself a breaking change for anyone who happened to call it, however unlikely that is. This is not
a cleanup that can ride along in an unrelated PR: it needs a version decision and a line in the
release notes.

## Scope

1. Decide for each of the two whether it was ever meant to be callable from outside.
   `truncateNumberToMaxDecimalDigits` carries a KDoc with worked examples, which reads like a
   deliberate helper; `firstChar` reads like a private convenience.
2. For whichever is not: make it `internal`. Unit tests live in the same module and still see it, so
   `CurrencyRounderTest` keeps working unchanged.
3. Run `./gradlew :library:apiDump` and commit the shrunken `library/api/library.api`. The removed
   lines are the change, and by the rule added in #20 they get a "Public API change" bullet in the
   PR description.
4. Pick the release increment accordingly — this is not a patch.

## Done when

- Every remaining entry in `library/api/library.api` is there because someone decided it should be.
- `./gradlew build` is green and the dump diff shows only the intended removals.

## References

- #20 — where the dump was introduced and this was spotted.

PR #24, opened 2026-09-06
Clean up the formatting core (#17)

### LINKED ISSUE #17: Clean up the formatting core: dead code, Array<Any?>, oversized onTextModified
## Why

Four findings in the formatting core, all in code that changes often. None of them is a bug today; all of them make the next bug harder to see. This is the only issue of the set that touches production code, so it goes after the gates in #15 are in place — detekt is what will keep `onTextModified` from growing back.

## Findings

1. **Commented-out code with a date** — `library/src/main/java/ru/pravbeseda/currencyedittext/watchers/CurrencyInputWatcher.kt:160-171`, the leading-zero removal "removed 23.06.2024". Git already holds it; the comment only makes the function longer.
2. **`Array<Any?>` as a return type** — `CurrencyInputWatcher.kt:238`, whose result is cast back at `:225` with `(text as? String?) ?: ""` and `(position as? Int?) ?: 0`. A `data class` with two fields removes both casts and the two silent fallbacks that hide a type error as an empty string and a zero cursor.
3. **Dead public API** — `truncateNumberToMaxDecimalDigits` (`library/src/main/java/ru/pravbeseda/currencyedittext/util/CurrencyRounder.kt:31`) is public, is not called anywhere in the library, and is exercised only by `CurrencyRounderTest`. Either the library should use it or it should go; keeping it means supporting it forever in a Maven Central artifact. Removing it is a public API change and belongs in a minor release, not a patch.
4. **`onTextModified` is 145 lines** (`CurrencyInputWatcher.kt:51`) with ~25 branches and three mutable positional variables (`resultText`, `position`, `cursorPosition`) rewritten by every step. Extract the steps that are already marked by comments — separator normalisation, sign handling, currency-prefix restoration, decimal-part trimming — into named private functions.

Separately noted, not part of this issue: `parseMoneyValueWithLocale` swallows `ParseException` and returns `0` (`util/Utils.kt:49`), so invalid input is indistinguishable from a real zero. Changing that changes behaviour for consumers, so it needs its own decision.

## How

Refactoring only — no behaviour change. The existing suite (`CurrencyInputWatcherTest`, 43 test methods across 5 locales) must stay untouched and green through every step: if a test needs editing, the change was not a refactoring. One finding per commit, smallest first.

## Done when

The four findings are gone, `./gradlew :library:test` is green with an unmodified test suite, and detekt no longer reports `LongMethod`/`CyclomaticComplexMethod` on `onTextModified`.

pravbeseda:
Closes #17. Refactoring only: `CurrencyInputWatcherTest` is untouched and its 43 methods across 5 locales stay green through every commit.

## What changed

- **Commented-out code** — the leading-zero removal disabled on 23.06.2024 is gone from `CurrencyInputWatcher.kt`; git history already holds it.
- **`Array<Any?>`** — `calculateSpacing` returned an array whose two elements `setText` cast back with `as? String? ?: ""` and `as? Int? ?: 0`, so a type error would have shown up as an empty field and a zero cursor. It returns a `TextWithCursor` now, which also let both of its parameters become non-null: every caller already passed a value.
- **Dead public API** — `truncateNumberToMaxDecimalDigits`, `util/CurrencyRounder.kt` and `CurrencyRounderTest` are removed.
- **`onTextModified`** — was ~145 lines with ~25 branches and three positional variables rewritten by every step. It is 11 lines now, calling one named function per step, each taking and returning a `TextWithCursor` instead of mutating shared state: `normalizeInput` → `extractSign` → `restoreCurrencyPrefix` → `keepLastDecimalSeparator` → `buildNumber` → `composeResult`.

## Public API change

`ru.pravbeseda.currencyedittext.util.CurrencyRounderKt.truncateNumberToMaxDecimalDigits` is removed. The `library/api/library.api` diff is in the same commit as the deletion. **The next release is 2.0.0, not a patch or a minor.** This also settles half of #22; `firstChar` is still open there.

## Why the formatter is its own class

Extracting the steps as private functions of `CurrencyInputWatcher` — what #17 asked for — pushed the class to 14 functions against detekt's `TooManyFunctions` threshold of 11, and the rule against new baseline entries left no way to silence it. The split it forced is the right one: `CurrencyTextFormatter` is `internal`, holds no reference to a view and is unit-testable on its own, while `CurrencyInputWatcher` is now the `EditText` adapter and nothing else. CLAUDE.md's Architecture section is updated to match.

## One behaviour difference

`cursorPositionWithoutGrouping` uses `take(end)` where the original used `substring(0, end)`. A negative cursor position still throws — `String.take` opens with `require(n >= 0)` — so the only thing that changes is the exception class: `IllegalArgumentException` where the old code threw `StringIndexOutOfBoundsException`. The same inputs throw on both sides, and neither path is covered by a test.

The crash behind it is not introduced here: it reproduces on `main` from the same gesture and is tracked in #26.

## Tests

No new tests: every change is a pure move or a mechanical rewrite of existing logic, and the existing suite pins the behaviour it must preserve — a test that had to be edited would have proved the change was not a refactoring. Removing `CurrencyRounderTest` is the deletion of the removed function's only call site, not a change to the watcher suite.

## Gates

`./gradlew build` is green. `:library:assembleDebugAndroidTest` compiles. Coverage 85.5% against the 80 bound. The detekt baseline goes from 25 entries to 14 — `LongMethod` and `CyclomaticComplexMethod` on `onTextModified`, `CyclomaticComplexMethod` and `NestedBlockDepth` on `calculateSpacing`, five `UseOrEmpty`, one `MagicNumber` and one `UnusedPrivateProperty` are gone. No entry was added.

## Follow-up

#23 — `parseMoneyValueWithLocale` swallows `ParseException` and returns `0`, which #17 listed and deliberately deferred.

https://claude.ai/code/session_014MHLrtD5cgPVkKGjcgyRi3


pravbeseda (commented):
🤖 Reviewed the whole PR; the refactoring checks out as behaviour-preserving (old and new algorithm run side by side over ~38k simulated edits diverge nowhere). 0 blocking findings, 1 suggestion follows.

library/src/main/java/ru/pravbeseda/currencyedittext/watchers/CurrencyTextFormatter.kt
```diff
@@ -0,0 +1,249 @@
...
+        return TextWithCursor(
+            integerPart + decimalSeparator + padFractionalPart(fractionalPart),
+            cursorPosition,
+        )
+    }
+
+    /** Cursor position in the text the grouping separators have been stripped from. */
+    private fun cursorPositionWithoutGrouping(input: TextWithCursor): Int {
+        val end = minOf(input.cursorPosition, input.text.length)
+        val separatorsBeforeCursor = input.text.take(end).count { it == groupingSeparator }
```
pravbeseda: 🤖 suggestion: The description names this line as the one behaviour difference — `take(end)` returning an empty string where `substring(0, end)` used to throw. I could not reproduce that. `String.take` opens with `require(n >= 0) { "Requested character count $n is less than zero." }`, so a negative `end` throws `IllegalArgumentException` instead of returning `""`. Running main's `onTextModified` and this formatter side by side over 200k random inputs, the two throw on exactly the same 37193 of them and only the exception class differs.

The path is also reachable from a real gesture rather than only a synthetic one: with `negativeValueAllow = true` and the default empty `currencySymbol`, a field showing `-5` with the caret at 0 and the user typing `-` gives `editPosition = 1`; `extractSign` takes it to `-1` (two minuses, `position > currencySymbol.length`) and nothing clamps it before it arrives here. main throws `StringIndexOutOfBoundsException` on that same input, so nothing has got worse — but should the description still say the crash is gone?

pravbeseda: Correct, and the description was wrong. `String.take` is `require(n >= 0) { ... }` then `substring(0, n.coerceAtMost(length))` (kotlin-stdlib 2.1.10, `commonMain/generated/_Strings.kt:570`), so a negative `end` throws `IllegalArgumentException` rather than returning `""`.

The description now says what actually changes — the exception class, on inputs that throw on both sides — and points at #26 for the crash itself, which reproduces on `main` from the same gesture and is not introduced here. No code change: the line behaves as intended for every reachable non-negative position.

PR #28, opened 2026-09-06
fix: two crashes on a plain keystroke in the formatter

### LINKED ISSUE #26: Typing a minus at the head of a negative value crashes when no currency symbol is set
Typing a second minus at the head of a negative value crashes the watcher.

## Reproduction

Configure the watcher with `negativeValueAllow = true` and the default `currencySymbol = ""`:

```kotlin
val config = CurrencyInputWatcherConfig(
    locale = Locale.US,
    currencySymbol = "",
    negativeValueAllow = true,
)
CurrencyTextFormatter(config, '.', ',').format(
    newPartOfText = "-",   // the user types a minus
    newText = "--5",       // the field showed "-5", the caret sat at 0
    oldText = "-5",
    editPosition = 1,      // EasyTextWatcher reports start = 0, count = 1
)
```

On `main` (pre-#24) this throws:

```
java.lang.StringIndexOutOfBoundsException: Range [0, -1) out of bounds for length 1
```

after #24 the same input throws:

```
java.lang.IllegalArgumentException: Requested character count -1 is less than zero.
```

Both propagate out of `TextWatcher.afterTextChanged`, so the app crashes on a plain keystroke.

## Cause

`CurrencyInputWatcher.kt:95` on `main` (`CurrencyTextFormatter.extractSign` after #24) subtracts 2 from
the cursor position when the text holds more than one minus:

```kotlin
if (numberOfMinus > 1 && position > config.currencySymbol.length) {
    position -= 2
}
```

With an empty currency symbol the guard is only `position > 0`, so `position = 1` becomes `-1`.
Nothing clamps it back: the `if (position < 0) position = 0` in the "leave last decimalSeparator
only" step runs only when the text contains a decimal separator, and `"--5"` does not. The negative
value then reaches the cursor recalculation — `CurrencyInputWatcher.kt:135` on `main`,
`CurrencyTextFormatter.kt:176` after #24 — which slices the string with it.

A non-empty currency symbol hides the bug: `"$ --5"` gives `position = 3` and the guard
`position > 2` still leaves `1`, which is non-negative.

## Notes

- Pre-existing on `main`; #24 only changes which exception is thrown, so this is not a regression
  introduced by that PR.
- Found by running `main`'s `onTextModified` and #24's `CurrencyTextFormatter` side by side over
  200k random inputs: the two throw on exactly the same 37193 of them.
- Not covered by `CurrencyInputWatcherTest`.

### LINKED ISSUE #27: Typing a decimal separator over a single zero crashes when maxNumberOfDecimalPlaces = 0
Typing a decimal separator into a field showing a single `0` crashes the watcher when `maxNumberOfDecimalPlaces = 0`.

## Reproduction

```kotlin
val config = CurrencyInputWatcherConfig(
    locale = Locale.US,
    currencySymbol = "",
    maxNumberOfDecimalPlaces = 0,
)
CurrencyTextFormatter(config, '.', ',').format(
    newPartOfText = ".",  // the user types a decimal separator
    newText = "0.",       // the field showed "0", the caret sat at the end
    oldText = "0",
    editPosition = 2,
)
```

throws

```
java.lang.IndexOutOfBoundsException: Range [2, 1) out of bounds for length 1
```

out of `TextWatcher.afterTextChanged`, so the app crashes on a plain keystroke. It reproduces with
`currencySymbol = "$ "` as well (`oldText = "$ 0"`, `newText = "$ 0."`, `editPosition = 4`), and with
`negativeValueAllow` either way — the flag plays no part.

## Cause

Two steps of `normalizeInput` disagree about the text length:

- `CurrencyTextFormatter.kt:68` ("Replace single zero to inserted digit") drops the `0`, so `"0."`
  becomes `"."` — but leaves `position` at 2;
- `CurrencyTextFormatter.kt:84` ("Remove the typed decimal separator when no decimal places are
  allowed") then calls `text.replaceRange(position - 1, position, "")`, i.e. `replaceRange(1, 2)` on
  a one-character string.

The first step assumes the inserted character is a digit; when it is the decimal separator and
`maxNumberOfDecimalPlaces = 0`, the separator is the only thing left in the text.

## Notes

- Pre-existing; unrelated to #26, which is about a second minus and is fixed separately.
- Found by fuzzing `CurrencyTextFormatter.format` over 200k random inputs (4 config combinations)
  after the #26 fix: this is the only crash class left.
- Not covered by `CurrencyInputWatcherTest`.

pravbeseda:
Two crashes on a plain keystroke, both inside `CurrencyTextFormatter.normalizeInput` /
`extractSign`, both pre-existing and neither covered by a test.

## #26 — a second minus at the head of a negative value

`extractSign` subtracts 2 from the cursor position for the two minuses it strips, guarded only by
`position > currencySymbol.length`. With an empty currency symbol that guard is `position > 0`, so
`position = 1` becomes `-1`, and the negative position later slices the string:
`IllegalArgumentException: Requested character count -1 is less than zero`. The step now clamps its
result at 0, which lands the cursor where the non-empty-symbol case already put it — right before
the first digit.

## #27 — a decimal separator over a single zero, with `maxNumberOfDecimalPlaces = 0`

The "replace single zero to inserted digit" step fired for any one-character insert, so typing `.`
over `"0"` dropped the zero and left the cursor two characters into a one-character string. The
next step, which removes a separator that is not allowed, then sliced at that stale position:
`IndexOutOfBoundsException: Range [2, 1) out of bounds for length 1`. The step now fires only for
the digit it is named after.

Side effect worth knowing: inserting a *non-digit* over a lone `0` (a letter, a space) now keeps the
zero instead of clearing the field. No existing test covered that; the new behaviour is the correct
one.

## Testing

- Both fixes are test-first: each test was watched failing with the exact exception from its issue
  before the fix, and both loop over the five locales of the suite, including `ru-Ru` with its empty
  currency symbol.
- `./gradlew build` passes locally — 44 unit tests, all six gates.
- Fuzzing `CurrencyTextFormatter.format` over 200k random inputs across four config combinations
  (currency symbol on/off × negative values on/off × 0 and 2 decimal places) reported 4 crash
  classes before and 0 after. The fuzz harness was a throwaway and is not part of the diff.

## Public API change

None — `CurrencyTextFormatter` is `internal`, `apiCheck` is unchanged.

Fixes #26
Fixes #27

### DIFF
--- a/library/src/main/java/ru/pravbeseda/currencyedittext/watchers/CurrencyTextFormatter.kt
+++ b/library/src/main/java/ru/pravbeseda/currencyedittext/watchers/CurrencyTextFormatter.kt
@@ -66,9 +66,11 @@ internal class CurrencyTextFormatter(
             text = text.replaceRange(position - 1, position, decimalSeparator.toString())
         }
 
-        // Replace single zero to inserted digit
+        // Replace the single zero with what was typed over it, unless that is a decimal separator:
+        // then the zero is the integer part of the number being entered
         if (oldText?.removePrefix(config.currencySymbol) == "0" &&
-            newPartOfText?.singleOrNull()?.isDigit() == true
+            newPartOfText?.length == 1 &&
+            !separatorTyped
         ) {
             text = text.replaceFirst("0", "")
         }
--- a/library/src/main/java/ru/pravbeseda/currencyedittext/watchers/CurrencyTextFormatter.kt
+++ b/library/src/main/java/ru/pravbeseda/currencyedittext/watchers/CurrencyTextFormatter.kt
@@ -67,7 +67,9 @@ internal class CurrencyTextFormatter(
         }
 
         // Replace single zero to inserted digit
-        if (oldText?.removePrefix(config.currencySymbol) == "0" && newPartOfText?.length == 1) {
+        if (oldText?.removePrefix(config.currencySymbol) == "0" &&
+            newPartOfText?.singleOrNull()?.isDigit() == true
+        ) {
             text = text.replaceFirst("0", "")
         }
 
--- a/library/src/main/java/ru/pravbeseda/currencyedittext/watchers/CurrencyTextFormatter.kt
+++ b/library/src/main/java/ru/pravbeseda/currencyedittext/watchers/CurrencyTextFormatter.kt
@@ -103,7 +103,7 @@ internal class CurrencyTextFormatter(
             position -= numberOfMinus
         }
 
-        return SignedText(sign, TextWithCursor(input.text.replace("-", ""), position))
+        return SignedText(sign, TextWithCursor(input.text.replace("-", ""), maxOf(0, position)))
     }
 
     /** Puts the currency symbol back when the user has deleted part of it. */
--- a/library/src/test/java/ru/pravbeseda/currencyedittext/CurrencyInputWatcherTest.kt
+++ b/library/src/test/java/ru/pravbeseda/currencyedittext/CurrencyInputWatcherTest.kt
@@ -673,6 +673,31 @@ class CurrencyInputWatcherTest {
     }
 
     @Test
+    fun `Issue #27 - typing a minus over a single zero must replace the zero`() {
+        for (locale in locales) {
+            val currentEditTextContent = "${locale.currencySymbol}0"
+            val expectedText = "${locale.currencySymbol}-"
+            val expectedCursorPosition = expectedText.length
+
+            val (editText, editable, watcher) = setupTestVariables(locale, 2, true)
+            `when`(editable.toString()).thenReturn("$currentEditTextContent-")
+
+            val start = currentEditTextContent.length
+            watcher.beforeTextChanged(
+                currentEditTextContent,
+                start,
+                0,
+                1,
+            )
+            watcher.onTextChanged(editable, start, 0, 1)
+            watcher.afterTextChanged(editable)
+
+            verify(editText, times(1)).setText(expectedText)
+            verify(editText).setSelection(expectedCursorPosition)
+        }
+    }
+
+    @Test
     fun `should remove any non digit character`() {
         for (locale in locales) {
             val currentEditTextContent = "- 10006metres"
--- a/library/src/test/java/ru/pravbeseda/currencyedittext/CurrencyInputWatcherTest.kt
+++ b/library/src/test/java/ru/pravbeseda/currencyedittext/CurrencyInputWatcherTest.kt
@@ -648,6 +648,31 @@ class CurrencyInputWatcherTest {
     }
 
     @Test
+    fun `Issue #27 - typing a decimal separator over a single zero must not crash when no decimal places are allowed`() {
+        for (locale in locales) {
+            val currentEditTextContent = "${locale.currencySymbol}0"
+            val expectedText = "${locale.currencySymbol}0"
+            val expectedCursorPosition = expectedText.length
+
+            val (editText, editable, watcher) = setupTestVariables(locale, 0)
+            `when`(editable.toString()).thenReturn("$currentEditTextContent${locale.decimalSeparator}")
+
+            val start = currentEditTextContent.length
+            watcher.beforeTextChanged(
+                currentEditTextContent,
+                start,
+                0,
+                1,
+            )
+            watcher.onTextChanged(editable, start, 0, 1)
+            watcher.afterTextChanged(editable)
+
+            verify(editText, times(1)).setText(expectedText)
+            verify(editText).setSelection(expectedCursorPosition)
+        }
+    }
+
+    @Test
     fun `should remove any non digit character`() {
         for (locale in locales) {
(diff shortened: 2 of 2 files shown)


library/src/main/java/ru/pravbeseda/currencyedittext/watchers/CurrencyTextFormatter.kt
```diff
@@ -67,7 +67,9 @@ internal class CurrencyTextFormatter(
         }
 
         // Replace single zero to inserted digit
-        if (oldText?.removePrefix(config.currencySymbol) == "0" && newPartOfText?.length == 1) {
+        if (oldText?.removePrefix(config.currencySymbol) == "0" &&
+            newPartOfText?.singleOrNull()?.isDigit() == true
```
pravbeseda: 🤖 blocking: I am not sure the "non-digit over a lone 0 keeps the zero" side effect is harmless for the minus sign. With `negativeValueAllow = true`, a field showing `$ 0` and the user typing `-` at the end: on `main` the lone zero was dropped and the field became `$ -`, and the next `5` gave `$ -5`. With this condition the zero survives, the field becomes `$ -0`, and the next `5` gives `$ -05` (same without a currency symbol: `-0` then `-05`), because the single-zero replacement only matches an `oldText` of exactly `0`, never `-0`. I checked both paths by running `CurrencyTextFormatter.format` on this commit and on `main`. Isn't that a regression on the most ordinary way of entering a negative amount over a zero — and shouldn't a test pin whichever result is intended here?

pravbeseda (commented):
🤖 Reviewed the two formatter fixes and their tests at 3c32501: 1 blocking finding, 0 suggestions.

library/src/main/java/ru/pravbeseda/currencyedittext/watchers/CurrencyTextFormatter.kt
pravbeseda: 🤖 blocking: I’m not sure the non-digit branch preserves the single-zero placeholder when negative input is enabled. Starting from `0`, inserting `-` before the zero formats the field as `-0` with the caret between the sign and zero; typing `5` next then produces `-50` instead of `-5` because this condition no longer removes the placeholder. Would this make that ordinary negative-number entry sequence change the value by a factor of ten?

pravbeseda (commented):
🤖 Reviewed the two formatter crash fixes; 1 blocking finding and 0 suggestions follow.

library/src/main/java/ru/pravbeseda/currencyedittext/watchers/CurrencyTextFormatter.kt
pravbeseda: Confirmed and fixed. Two independent reviews raised this same regression on this line; it is answered once here.

Reproduced by running `CurrencyInputWatcher` on this commit and on `main` (`negativeValueAllow = true`, `en-US`, symbol `"$ "`):

| sequence | `main` | this commit before the fix |
|---|---|---|
| `$ 0`, caret before the zero, `-` then `5` | `$ -` → `$ -5` | `$ -0` → `$ -50` |
| `$ 0`, caret at the end, `-` then `5` | `$ -` → `$ -5` | `$ -0` → `$ -05` |

Both descriptions were right, each for its caret position.

The condition now names the case the #27 fix was actually about instead of the character class:

```kotlin
if (oldText?.removePrefix(config.currencySymbol) == "0" &&
    newPartOfText?.length == 1 &&
    !separatorTyped
) {
```

So the lone zero survives only when a decimal separator is typed over it — where it is the integer part of the number — and is replaced by anything else, as on `main`. A test pinning the minus case was added first and watched fail (`setText("$ -0")` where `"$ -"` was wanted); it is green now, together with the whole suite and `./gradlew build`.

pravbeseda: Confirmed and fixed. Two independent reviews raised this same regression on this line; it is answered once here.

Reproduced by running `CurrencyInputWatcher` on this commit and on `main` (`negativeValueAllow = true`, `en-US`, symbol `"$ "`):

| sequence | `main` | this commit before the fix |
|---|---|---|
| `$ 0`, caret before the zero, `-` then `5` | `$ -` → `$ -5` | `$ -0` → `$ -50` |
| `$ 0`, caret at the end, `-` then `5` | `$ -` → `$ -5` | `$ -0` → `$ -05` |

Both descriptions were right, each for its caret position.

The condition now names the case the #27 fix was actually about instead of the character class:

```kotlin
if (oldText?.removePrefix(config.currencySymbol) == "0" &&
    newPartOfText?.length == 1 &&
    !separatorTyped
) {
```

So the lone zero survives only when a decimal separator is typed over it — where it is the integer part of the number — and is replaced by anything else, as on `main`. A test pinning the minus case was added first and watched fail (`setText("$ -0")` where `"$ -"` was wanted); it is green now, together with the whole suite and `./gradlew build`.

Issue #21, opened 2026-09-06
Upgrade Kotlin and move ABI validation to the Kotlin plugin's built-in abiValidation {}


pravbeseda:
## Why

ABI validation landed in #20 on the standalone `org.jetbrains.kotlinx.binary-compatibility-validator`
plugin (`gradle/libs.versions.toml:18`, applied at `library/build.gradle:6`). That was the only
option available: the Kotlin Gradle Plugin's own `abiValidation {}` needs Kotlin 2.2+, and this
project is on Kotlin 2.1.10 (`gradle/libs.versions.toml:11`).

The built-in one is where upstream is going — the standalone plugin is in maintenance and the
feature now lives inside KGP. Staying on the standalone plugin means carrying a separate dependency
and a separate dump format for a job the toolchain will do itself. Kotlin 2.1.10 is also two
minor lines behind: 2.4.10 is the current stable release.

## Scope

Two steps that have to happen in this order, and they may be worth two pull requests:

1. **Upgrade Kotlin.** 2.1.10 → 2.4.10 (or whatever is stable at the time). Check the KGP ↔ Gradle
   support matrix first: the wrapper is on Gradle 8.13
   (`gradle/wrapper/gradle-wrapper.properties:3`) and AGP on 8.13.0, so this may drag a Gradle bump
   with it. `jvmTarget = JVM_1_8` is set in both modules (`library/build.gradle:39`,
   `sample/build.gradle:30`) and must stay: Java source/target is 1.8 and minSdk is 19.
2. **Swap the plugin.** Remove the standalone plugin and its catalog entry, enable
   `abiValidation {}` in `:library`, regenerate the dump, and update the CI step, which today runs
   `./gradlew :library:apiCheck` (`.github/workflows/ci.yml:108`). The built-in feature uses its own
   task names and its own dump location, so the CI step, the `CLAUDE.md` gates table and the
   `apiDump` instruction in `CLAUDE.md` all change with it.

## Known complication: detekt

detekt 1.23.8 is compiled against Kotlin 2.0.21 and already prints a version-mismatch warning on
2.1.10 — see the comment in the root `build.gradle` plugins block. Raising Kotlin widens that gap,
and detekt 2.x wants Gradle 8.14.3+. So the Kotlin upgrade forces a decision on detekt rather than
being independent of it. Decide it deliberately, do not let the warning turn into a silently broken
gate: verify after the upgrade that `./gradlew detektAll` still analyses with type resolution and
[...]

PR #29, opened 2026-09-07
chore: upgrade Kotlin to 2.4.10 and Gradle to 8.14.5

pravbeseda:
First of the two changes issue #21 asks for. It does the prerequisite only: the
Kotlin Gradle Plugin's built-in `abiValidation {}` needs Kotlin 2.2+, and this
project was on 2.1.10. The plugin swap itself follows in its own pull request,
so the API dump's format change arrives as a diff with nothing else in it.

## What changed

- **Kotlin 2.1.10 → 2.4.10.** 2.4.10 is the current stable line; 2.4.20 is still
  an RC.
- **Gradle 8.13 → 8.14.5.** Not cosmetic: Kotlin 2.4.10 runs on 8.13 but warns
  that it is deprecated and that Kotlin 2.5.0 will require 8.14.4+. Clearing it
  now also lifts the Gradle floor detekt 2.x will ask for. Gradle 9.x is a
  separate job — the build still reports pre-existing Gradle 9 deprecations.
- **`CurrencyEditTextTest.kt`:** `arrayOf(BigDecimal(...), "...")` inferred its
  reified type argument as an intersection type, which Kotlin 2.4 rejects where
  2.1 only warned. Replaced with a `Pair`, which says the same thing with static
  types and drops both `as` casts. This was the only compilation failure of the
  upgrade.
- **AGP stays at 8.13.0** and detekt at 1.23.8 — see below.
- Comments that stated the old versions were updated: the detekt note in the
  root `build.gradle`, the two quality-gate notes in `gradle/libs.versions.toml`,
  and the toolchain line in `CLAUDE.md`.
- `.github/tasks/21-kotlin-upgrade-and-builtin-abi-validation.md` records the
  plan for both pull requests and the version facts it was decided on.

## Public API change

None. `./gradlew :library:apiDump` rewrites `library/api/library.api` byte for
byte — `git diff --exit-code` on it is clean, so the upgrade moved no signature.
The file is untouched in this diff, which is the point of checking it here
rather than after the plugin swap.

## detekt

detekt stays at 1.23.8. 2.x has been in alpha since 2025 and lives under a
different group id (`dev.detekt`); an alpha is not what a build gate is made of.
It is compiled against Kotlin 2.0.21, now four minor lines behind, so its typed
run was proven rather than assumed: a deliberate `UnsafeCallOnNullableType` was
planted in `library/src/main` and `./gradlew detektAll` failed the build on it,
which it can only do with type resolution working. The probe was removed.

## Verification

- `./gradlew build` — green, all six gates ran. Filtered line coverage 85.6%
  against the `minBound(80)` floor.
- The KGP "Gradle 8.13 is deprecated" warning is gone after the wrapper bump.
- `jvmTarget = JVM_1_8` and Java source/target 1.8 are untouched, and the build
  reports no JVM-target mismatch.
- `core.autocrlf` is `input` and there is no `.gitattributes`, so the
  regenerated `gradlew.bat` is stored with LF exactly as before.

## Tests

No new tests. `CLAUDE.md` exempts build-script changes, and the one source edit
is a fix to an existing instrumented test, not new logic in `library/src/main`.


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,7 @@
-#Wed Sep 07 10:27:56 MSK 2022
 distributionBase=GRADLE_USER_HOME
-distributionUrl=https\://services.gradle.org/distributions/gradle-8.13-bin.zip
 distributionPath=wrapper/dists
-zipStorePath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.14.5-bin.zip
```
pravbeseda: 🤖 suggestion: The distribution moves to 8.14.5 here, but the committed wrapper JAR has Gradle 8.13's published checksum (`81a82aa…`, versus `7d3a4ac…` for 8.14.5), and the launcher scripts retain the 8.13 invocation shape. I am not sure the wrapper itself was upgraded along with the distribution — could this leave two wrapper generations in the same bootstrap?

pravbeseda (commented):
🤖 Reviewed the Kotlin and Gradle toolchain upgrade; 0 blocking findings and 1 suggestion follow.

gradle/wrapper/gradle-wrapper.properties
pravbeseda: Confirmed and fixed. `sha256(gradle/wrapper/gradle-wrapper.jar)` was `81a82aa…`, which is the checksum Gradle publishes for the **8.13** wrapper jar, while `distributionUrl` already pointed at 8.14.5.

Cause: `./gradlew wrapper --gradle-version 8.14.5` writes the wrapper files using the Gradle that is *currently running* — 8.13 at that moment — so the first run only rewrites `distributionUrl`, and the jar plus the launcher scripts stay one generation behind. The documented remedy is to run the task a second time, now that the wrapper itself boots 8.14.5.

Done, and verified against the published checksum rather than by eye:

```
sha256(gradle/wrapper/gradle-wrapper.jar) = 7d3a4ac4de1c32b59bc6a4eb8ecb8e612ccd0cf1ae1e99f66902da64df296172
https://services.gradle.org/distributions/gradle-8.14.5-wrapper.jar.sha256
                                         = 7d3a4ac4de1c32b59bc6a4eb8ecb8e612ccd0cf1ae1e99f66902da64df296172
```

`gradlew` and `gradlew.bat` were regenerated with it, and `./gradlew build` is green on all six gates.

Issue #31, opened 2026-09-07
Instrumented test shouldSetDecimalZerosPadding depends on the device's default locale


pravbeseda:
## What happens

`CurrencyMaterialEditTextTest.shouldSetDecimalZerosPadding` never calls `setSeparators` or
`setLocale`, so it inherits whatever separators the device's default locale gives. On a device
whose locale uses `,` as the decimal separator it fails:

```
org.junit.ComparisonFailure: expected:<1[00.10]> but was:<1[ 001]>
  at ru.pravbeseda.currencyedittext.CurrencyMaterialEditTextTest.testSetText(CurrencyMaterialEditTextTest.kt:88)
  at ru.pravbeseda.currencyedittext.CurrencyMaterialEditTextTest.shouldSetDecimalZerosPadding(CurrencyMaterialEditTextTest.kt:76)
```

`"100.1"` is read with `.` as a grouping separator, so the digits become `1001` and the field
shows `1 001`.

## Reproduction

On an emulator whose locale is `ru-RU` (API 24, `Medium_Phone_24-7.0`):

```bash
adb shell setprop persist.sys.locale ru-RU && adb reboot
./gradlew :library:connectedDebugAndroidTest
```

9 tests, 1 failure. The same run on the same emulator switched to `en-US` is green, as is API 36
with `en-US`.

## Where

`library/src/androidTest/java/ru/pravbeseda/currencyedittext/CurrencyMaterialEditTextTest.kt:72-77`
— `shouldSetDecimalZerosPadding` is the only test in either instrumented class that asserts on a
string containing a separator without first pinning the separators. `shouldSetText`,
`shouldSetLocale` and every test in `CurrencyEditTextTest` do pin them.

## Why it matters now

The instrumented tests are no longer compile-only: they run on every push to `main` and as a gate
in the `Release` workflow. GitHub's runners are `en-US`, so the suite is green there today — but a
test whose verdict depends on the machine's locale is a false red waiting for whoever runs it
locally, and a blocked release if a runner image ever changes.

## Fix

Call `setSeparators(' ', '.')` in `shouldSetDecimalZerosPadding` the way `shouldSetText` already
does.

PR #32, opened 2026-09-07
chore: raise minSdk to 24, sweep the dependencies, run the instrumented tests

### LINKED ISSUE #16: Add Renovate and update the 11 outdated dependencies
## Why

Dependencies are updated by hand, in occasional sweeps. Android Lint currently reports 11 outdated coordinates:

- `compileSdk 34` → 36
- `com.google.android.material:material 1.9.0` → 1.14.0
- `androidx.appcompat:appcompat 1.6.1` → 1.8.0
- `androidx.test:core-ktx 1.5.0` → 1.7.0, `androidx.test.ext:junit-ktx 1.1.5` → 1.3.0, `androidx.test:runner 1.5.2` → 1.7.0, `androidx.test:rules 1.5.0` → 1.7.0
- `org.mockito:mockito-core 5.20.0` → 5.23.0
- plus the `:sample` module's own androidx versions

`material` is an `api` dependency of a published library, so its version is part of this library's contract for every consumer.

## Scope

- [ ] `renovate.json5`: weekly schedule, `config:recommended` + vulnerability alerts, `dependencyDashboard`, patch updates automerged, minor and major read by a person, androidx grouped, Kotlin toolchain never automerged.
- [ ] Work through the 11 outdated coordinates above. `material 1.9 → 1.14` and `compileSdk 34 → 36` change behaviour: run the `:sample` app on a device and check both components before merging.

Depends on the version catalog from #15 — Renovate should have one file to open pull requests against.

## Done when

Renovate opens its first dashboard issue, and the dependency list carries no lint warning that was not deliberately deferred with a reason.

### LINKED ISSUE #18: Deferred quality work: instrumented tests in CI, dependency analysis, ABI validation

pravbeseda:
Closes #16. Closes #18.

## Public API change

`minSdk` rises from 19 to 24. `material` is an `api` dependency of the published artifact, so its
requirements are this library's requirements: a consumer on API 19–22 will no longer be able to
build against it. `library/api/library.api` is byte-for-byte unchanged — no signature moved — but
the contract did, so **the next release is a minor bump, not a patch**. The README now states the
requirement, and names 1.1.1 as the last release supporting `minSdk 19`.

## Why 24 and not 23

The dependencies demand 23: `material 1.14.0` and `appcompat 1.8.0` both refuse a lower `minSdk`,
`constraintlayout 2.2.2` refuses below 21. 24 was chosen deliberately over the bare minimum.

`androidx.core` stops at 1.17.0: 1.18.0 and above require Android Gradle plugin 9.1.0. Renovate
will keep offering the newer one; that pull request waits for the AGP upgrade, and CLAUDE.md says
so.

## What is in the diff

**#16 — Renovate and the dependency sweep**

- `compileSdk`/`targetSdk` 34 → 37 in both modules, `minSdk` 19 → 24.
- `appcompat 1.6.1 → 1.8.0`, `constraintlayout 2.1.4 → 2.2.2`, `core-ktx 1.10.1 → 1.17.0`,
  `material 1.9.0 → 1.14.0`, `mockito 5.20.0 → 5.23.0`, `androidx.test` core/rules/runner
  `→ 1.7.0`, `ext:junit-ktx 1.1.5 → 1.3.0`.
- `renovate.json5`: weekly, `config:recommended` plus vulnerability alerts and the dependency
  dashboard, patch releases automerged, minor and major read by a person, androidx grouped, the
  Kotlin/AGP toolchain never automerged whatever the update type.

**#18 — the deferred quality work**

- The nine instrumented tests over `CurrencyEditText` and `CurrencyMaterialEditText` stop being
  compile-only. A new `instrumented` job in `ci.yml` runs them on an API 24 emulator on every push
  to `main` and on manual dispatch; `release.yml` runs the same step between `./gradlew build` and
  the irreversible `publishAndReleaseToMavenCentral`. They deliberately do **not** run on pull
  requests: the emulator costs about ten minutes and the two Views change rarely.
- `dependency-analysis` 3.19.1, wired as `./gradlew buildHealth` and kept out of `check` — it has
  no baseline, and "unused" is a false positive by construction for a dependency pulled in for its
  resources. It already demonstrates this: it reports `material` as unused in `:sample`, which is
  correct as declared.
- The third item of #18, binary compatibility validation, was already done in #20 and #30.

**Fallout of the `minSdk` bump, handled here rather than baselined**

`ObsoleteSdkInt` turned the API-level guards into lint errors. `isLollipopAndAbove()`, the two
`@RequiresApi(Build.VERSION_CODES.LOLLIPOP)` annotations and the `if (isLollipopAndAbove() && …)`
conditions are removed; `isApi26AndAbove()` stays, because 26 is still above 24.
`sample/src/main/res/drawable-v24` is merged into `drawable`. The `OldTargetApi` entry left both
`lint-baseline.xml` files — baselines only ever shrink.

`OldTargetApi` itself is demoted to `informational`, joining `GradleDependency` and
`NewerVersionAvailable`. It judges `targetSdk` against the newest SDK platform installed on the
machine, so `targetSdkVersion 36` was green locally and red on the runner; and it cannot be frozen
instead, because a baseline entry carries the `targetSdk` of the day it was written and goes stale
on the next bump. It stays in the lint report CI uploads.

AGP 8.13.0 prints a warning that it has not been tested against compile SDK 37. It is left visible
rather than silenced with `android.suppressUnsupportedCompileSdk`; the next AGP upgrade removes
it.

## Verification

- `./gradlew build` green: Spotless, detekt with type resolution, Kover (86.1%, floor 80), Android
  Lint with `warningsAsErrors`, unit tests, `checkKotlinAbi`.
- `./gradlew :library:connectedDebugAndroidTest` — 9/9 on an **API 24** emulator (the new `minSdk`)
  and 9/9 on **API 36**, both run locally. This is the real check on the whole upgrade: the two
  Views are unreachable from `src/test`, and `material 1.14` at `minSdk 24` is exactly the
  combination worth doubting.
- `renovate-config-validator` accepts `renovate.json5`.
- `./gradlew check` does not depend on any `buildHealth` task.

## No tests, because

Everything here is a build-script change plus the removal of guards that `minSdk 24` makes
unreachable — both named as exceptions in CLAUDE.md. The behaviour of the removed code on every
still-supported API level is unchanged, and the existing instrumented suite covers it on the
lowest one.

## Two things this pull request cannot do

- `renovate.json5` configures the bot; it does not enable it. The Renovate GitHub App has to be
  installed on the repository once, by hand, from https://github.com/apps/renovate — and
  "Allow auto-merge" turned on in the repository settings for the patch rule to work.
- Found on the way and filed rather than fixed here: #31 —
  `CurrencyMaterialEditTextTest.shouldSetDecimalZerosPadding` never pins the separators, so it
  fails on a device whose locale uses `,` as the decimal separator. Harmless while the tests were
  compile-only; now they gate a release.

Plan: `.github/tasks/16-18-dependencies-and-deferred-quality.md`.


renovate.json5
```diff
@@ -0,0 +1,59 @@
...
+      groupName: 'androidx',
+    },
+    {
+      // The toolchain decides whether everything else compiles at all, and each
+      // of these upgrades has already needed manual work in this repository
+      // (#21). Never automerged, whatever the update type.
+      matchPackageNames: [
+        'org.jetbrains.kotlin**',
+        'com.android.tools.build**',
+        'com.android.application',
```
pravbeseda: 🤖 blocking: I am not sure this rule ever matches the Android Gradle plugin. `matchPackageNames` is compared against Renovate's `packageName` (lib/util/package-rules/package-names.ts), and for a `[plugins]` entry of a version catalog the gradle manager sets `packageName` to `<id>:<id>.gradle.plugin` (lib/modules/manager/gradle/extract/catalog.ts) — so for `android-application` it is `com.android.application:com.android.application.gradle.plugin`. A bare `com.android.application` is an exact match and would not match that string, and `com.android.tools.build**` matches nothing declared in `libs.versions.toml` at all. The Kotlin line is fine because `org.jetbrains.kotlin**` covers the `:…gradle.plugin` suffix, and `gradle` is fine because the wrapper's `packageName` falls back to its `depName`.

If that reading is right, an AGP patch release (say 8.13.1) falls through to the `patch` rule above and is automerged — the one case this rule says never happens. Could you check this against Renovate's dependency dashboard or a dry run once the app is installed?

renovate.json5
```diff
@@ -0,0 +1,59 @@
...
+  $schema: 'https://docs.renovatebot.com/renovate-schema.json',
+  // Renovate reads gradle/libs.versions.toml natively, so every pull request it
+  // opens is a one-line diff in the one file a version is allowed to live in.
+  //
+  // This file only configures the bot; it does not enable it. The Renovate GitHub
+  // App has to be installed on the repository once, by hand, from
+  // https://github.com/apps/renovate.
+  extends: [
+    'config:recommended',
+    ':dependencyDashboard',
```
pravbeseda: 🤖 suggestion: `config:recommended` already extends `:dependencyDashboard` (it is the first entry of that preset's list), so this line adds nothing — unless it is meant as documentation, it could go.

renovate.json5
```diff
@@ -0,0 +1,59 @@
...
+      // "Allow auto-merge" enabled in the repository settings; without it
+      // Renovate leaves the pull request open and nothing is lost.
+      matchUpdateTypes: ['patch'],
+      automerge: true,
+    },
+    {
+      // Minor and major updates change behaviour, and `material` is an `api`
+      // dependency of a published artifact, so its version is part of this
+      // library's contract. A person reads these.
+      matchUpdateTypes: ['minor', 'major'],
```
pravbeseda: 🤖 suggestion: `automerge` defaults to `false`, and nothing earlier in this file turns it on for minor or major updates, so this rule cannot change any outcome. Is it kept on purpose as a statement of intent? If not, the intent already lives in the comment on the `patch` rule and this one could be dropped.

pravbeseda (commented):
🤖 Reviewed the whole PR (minSdk bump, dependency sweep, Renovate config, instrumented CI/release jobs, lint fallout): 1 blocking finding and 2 suggestions follow, all in renovate.json5.

renovate.json5
pravbeseda: Confirmed against Renovate's own source, and fixed.

- `lib/modules/manager/gradle/extract/catalog.ts:314` sets `packageName: `${depName}:${depName}.gradle.plugin`` for a `[plugins]` entry, so AGP arrives as `com.android.application:com.android.application.gradle.plugin`.
- `lib/util/package-rules/package-names.ts` matches `matchPackageNames` against `packageName` only — and returns `false` outright when it is absent, so there is no fall-back to `depName` at that point.
- `lib/workers/repository/process/fetch.ts:37` does `dep.packageName ??= dep.depName`, which is why the bare `gradle` entry works for the wrapper: the wrapper manager sets `depName: 'gradle'` and no `packageName`. It does not help the plugin entries, which set one explicitly.

So a bare `com.android.application` matched nothing and an AGP patch release would have fallen through to the `patch` rule and been automerged. `com.android.tools.build**` matched nothing declared in this repository either.

The three android entries are now one glob, `com.android.**`, which covers both the plugin-marker spelling and `com.android.tools.build:gradle` should AGP ever be declared as a library. `minimatch` is configured with `{ dot: true, nocase: true }` and the package name contains no `/`, so the trailing `**` matches the `:…gradle.plugin` suffix the same way `org.jetbrains.kotlin**` already did. The reasoning is now a comment on the rule, so the next person editing it does not have to rediscover it.

`renovate-config-validator` still accepts the file. Checking it against the dependency dashboard once the app is installed remains worth doing — the source says what it says, but a live dry run is the only thing that proves it.

renovate.json5
pravbeseda: Correct that it is redundant today, and kept anyway — the written approach is equally valid, so the incumbent stands.

`:dependencyDashboard` being inside `config:recommended` is a fact about that preset on the day it is read, not a promise. The explicit `extends` entry pins the dashboard to this repository whatever the preset's list becomes, and the dashboard is one of the things #16 asked for by name, so having it written keeps the config traceable to the issue that ordered it. That is worth one line.

renovate.json5
pravbeseda: Correct that it changes no outcome today, and kept anyway — the written approach is equally valid, so the incumbent stands.

The rule above this one turns automerge on. An explicit "not for minor and major" is what stops a later widening of that rule — a `matchUpdateTypes` grown by one entry, a preset that starts enabling automerge — from quietly taking the toolchain and `material` with it. Relying on a default to hold a policy that the same file overrides three lines earlier is the fragile half of the choice, not the safe one.

It also carries the reason `material` cannot be automerged at all, which is not derivable from the `patch` rule's comment: `material` is an `api` dependency of a published artifact, so its version is part of this library's contract.

PR #33, opened 2026-09-07
fix: pin the separators in shouldSetDecimalZerosPadding

### LINKED ISSUE #31: Instrumented test shouldSetDecimalZerosPadding depends on the device's default locale

pravbeseda:
Closes #31.

`CurrencyMaterialEditTextTest.shouldSetDecimalZerosPadding` asserted on `"100.10"` without pinning
the separators, so it inherited them from the device's default locale. On a ru-RU device `"100.1"`
was read with `.` as a grouping separator and the field showed `1 001`. It was the only test in
either instrumented class asserting on a string with a separator without setting one first.

The fix is the three lines the issue prescribes: `setSeparators(' ', '.')` on the main thread,
the way `shouldSetText` already does it.

## Verification

The emulator here is a production image, so `persist.sys.locale` cannot be set. Instead a throwaway
instrumented test pinned `Locale("ru", "RU")` explicitly and reproduced the issue's exact failure:

```
org.junit.ComparisonFailure: expected:<1[00.10]> but was:<1[ 001]>
```

The same test with `setSeparators(' ', '.')` passed. The scratch file was then deleted, and the real
suite runs green: `OK (9 tests)` on API 36, plus `./gradlew build` locally.

No production code changed, so no new unit test — this is a fix to a test.

https://claude.ai/code/session_01DKQTpTPQT9XphP1WAX8zV4

### DIFF
--- a/library/src/androidTest/java/ru/pravbeseda/currencyedittext/CurrencyMaterialEditTextTest.kt
+++ b/library/src/androidTest/java/ru/pravbeseda/currencyedittext/CurrencyMaterialEditTextTest.kt
@@ -72,6 +72,9 @@ class CurrencyMaterialEditTextTest {
 
     @Test
     fun shouldSetDecimalZerosPadding() {
+        InstrumentationRegistry.getInstrumentation().runOnMainSync {
+            currencyEditText.setSeparators(' ', '.')
+        }
         currencyEditText.setDecimalZerosPadding(true)
         testSetText("100.1", "100.10")
         currencyEditText.setDecimalZerosPadding(false)


library/src/androidTest/java/ru/pravbeseda/currencyedittext/CurrencyMaterialEditTextTest.kt
```diff
@@ -72,6 +72,9 @@ class CurrencyMaterialEditTextTest {
 
     @Test
     fun shouldSetDecimalZerosPadding() {
+        InstrumentationRegistry.getInstrumentation().runOnMainSync {
+            currencyEditText.setSeparators(' ', '.')
```
pravbeseda: 🤖 suggestion: this is now the same `runOnMainSync { setSeparators(' ', '.') }` block as in `shouldSetText` (lines 45-47), and the issue's point was that a test in this class inherits the device locale unless it pins the separators itself. Would pinning them once in `init()` instead cover both tests with one block — and any test added to this class later — rather than relying on each test remembering to do it? `shouldSetLocale` overrides them anyway, and `shouldNegativeValueAllow` has no separator in its samples, so I don't think either would notice. Your call — it's the fix the issue prescribes as it stands.

pravbeseda (commented):
🤖 Reviewed the three-line fix in CurrencyMaterialEditTextTest — 0 blocking findings, 1 suggestion.

library/src/androidTest/java/ru/pravbeseda/currencyedittext/CurrencyMaterialEditTextTest.kt
pravbeseda: Tried it, and the premise does not hold: `shouldSetLocale` does **not** override the separators.

`setLocale` (`CurrencyEditText.kt:126-131`) only replaces `locale`; `decimalSeparator` and
`groupingSeparator` keep whatever `setSeparators` put there (`CurrencyEditText.kt:142-155`), so an
explicit pin outranks the locale for the life of the view. With the block moved into `init()`,
`shouldSetLocale` fails on the emulator:

```
org.junit.ComparisonFailure: expected:<1[,]000.45> but was:<1[ ]000.45>
```

Keeping it in `init()` therefore means pinning the separators again inside `shouldSetLocale`, per
locale — the duplication moves rather than goes, and that test stops checking the one thing it
exists to check. So the block stays where the issue prescribes it, on the ground that the written
approach is equally valid: three repeated lines in a test that is explicit about its own
preconditions beat a shared `@Before` the locale test has to undo.

The suite is green as it stands: `OK (9 tests)` on API 36.

PR #35, opened 2026-09-07
refactor!: hide firstChar from the published API

### LINKED ISSUE #22: Review the public surface of util/: truncateNumberToMaxDecimalDigits and firstChar

pravbeseda:
Closes #22.

## What

`String?.firstChar()` (`library/src/main/java/ru/pravbeseda/currencyedittext/util/Routines.kt:57`)
is now `internal`. It is a one-line convenience for reading the first character of an XML attribute,
called only from the `init` blocks of the two Views (`CurrencyEditText.kt:84,86`,
`CurrencyMaterialEditText.kt:74,76`). Nothing in `:sample` uses it. It was public only because
nobody had said otherwise — unlike `emptyChar` three lines above it, which carries a comment
recording that its name is deliberate.

The issue's other half was already done: `truncateNumberToMaxDecimalDigits` was removed in 295fea1,
which is on `main` and not yet released.

## Public API change

`library/api/library.api` loses exactly one line:

```
-	public static final fun firstChar (Ljava/lang/String;)Ljava/lang/Character;
```

Regenerated with `./gradlew :library:updateKotlinAbi` (the issue names `apiDump`, the spelling from
the binary-compatibility-validator plugin that #21 replaced).

This narrows the surface published to Maven Central, so the removal is source- and
binary-incompatible for anyone who called it. Together with the already-merged removal of
`truncateNumberToMaxDecimalDigits` it goes out under the increment #24 already recorded for that
one: **the next release is 2.0.0**, not a patch and not a minor.

## Tests

No tests, because this changes visibility and not behaviour: the four call sites are unchanged and
live in the same module. The ABI dump diff is the verification.

## Checks

`./gradlew build` green locally — all six gates, `checkKotlinAbi` included.

### DIFF
--- a/library/src/main/java/ru/pravbeseda/currencyedittext/util/Routines.kt
+++ b/library/src/main/java/ru/pravbeseda/currencyedittext/util/Routines.kt
@@ -54,4 +54,4 @@ class Routines {
     }
 }
 
-fun String?.firstChar(): Char? = if (!this.isNullOrEmpty()) this[0] else null
+internal fun String?.firstChar(): Char? = if (!this.isNullOrEmpty()) this[0] else null
--- a/library/api/library.api
+++ b/library/api/library.api
@@ -131,7 +131,6 @@ public final class ru/pravbeseda/currencyedittext/util/Routines$Companion {
 
 public final class ru/pravbeseda/currencyedittext/util/RoutinesKt {
 	public static final field emptyChar C
-	public static final fun firstChar (Ljava/lang/String;)Ljava/lang/Character;
 }
 
 public final class ru/pravbeseda/currencyedittext/watchers/CurrencyInputWatcher : ru/pravbeseda/currencyedittext/watchers/EasyTextWatcher {


library/src/main/java/ru/pravbeseda/currencyedittext/util/Routines.kt
```diff
@@ -54,4 +54,4 @@ class Routines {
     }
 }
 
-fun String?.firstChar(): Char? = if (!this.isNullOrEmpty()) this[0] else null
+internal fun String?.firstChar(): Char? = if (!this.isNullOrEmpty()) this[0] else null
```
pravbeseda: 🤖 blocking: `firstChar` was already part of the 1.1.1 source API, and PR #24 explicitly records 2.0.0 for the other merged public-API removal, but this PR's Public API change section says the combined removals make a minor release. Would following that recorded release decision publish these source-incompatible removals as 1.2.0?

pravbeseda (commented):
🤖 Reviewed the whole pull request; 1 blocking finding and 0 suggestions follow.

library/src/main/java/ru/pravbeseda/currencyedittext/util/Routines.kt
pravbeseda: Correct, and the PR description was wrong. #24 records "The next release is 2.0.0, not a patch or a minor." for the `truncateNumberToMaxDecimalDigits` removal, and this PR removes a second published declaration on top of it, so calling the combination a minor would have published two source-incompatible removals as 1.2.0.

The description now says 2.0.0 and points at the decision in #24 instead of restating one of its own. No code change: the version is picked when the Release workflow is started, not in this branch.

PR #36, opened 2026-09-07
feat: calculator core — expression evaluator and key state machine

pravbeseda:
Groundwork for an optional calculator that both views will be able to open from a button at the
right edge of the field. This pull request adds the arithmetic and the key semantics only — no UI,
no new dependency, and no public API change: every class added here is `internal`.

The plan for both pull requests, including the survey of ready-made libraries that found nothing
usable, is in `.github/tasks/calculator-input-panel.md`.

## What is in it

- **`ExpressionEvaluator`** — recursive descent over `BigDecimal` for `+ - * / ( )` with a unary
  minus. Recursive descent rather than shunting-yard: precedence, parentheses and the unary minus
  all fall out of the grammar, so there is no tokeniser and no intermediate representation to get
  wrong. Division by zero and malformed input return `EvalResult.Failure`; nothing throws.
- **`CalculatorKey`** — the 20 keys of the panel. `symbol` is what a key appends; for `±`, `⌫` and
  `C` it is the label the panel will draw.
- **`CalculatorState`** — immutable state machine over a canonical ASCII expression. Holds no view
  reference and knows nothing about locale, so the panel translates the user's decimal separator
  on the way in and draws `×` and `÷` on the way out. A press that would make the expression
  unreadable — a second decimal point in one operand, a digit glued to a closing parenthesis — is
  ignored rather than accepted and left for the user to back out of.

## Two decisions worth a look

**Rounding.** Intermediate operations run at 16 significant digits, and only the final result is
rounded to the caller's scale, `HALF_UP`. Rounding every step to the field's scale instead would
make `100 ÷ 3 × 3` come out at 99.99, and the error would accumulate silently down a long
expression.

**The sign key never writes a double sign.** A unary `-` in front of the operand is removed; a
binary `+` or `-` in front of it is flipped instead, so `100-50` becomes `100+50` and not
`100--50`. Only `*` and `/`, which cannot absorb a sign, get a unary minus written after them.
The operation is reversible in both directions.

| Expression | `±` | Result |
|---|---|---|
| `100-50` | → | `100+50` |
| `100+50` | → | `100-50` |
| `100+` | → | `100-` |
| `100*3` | → | `100*-3` |
| `100` | → | `-100` |

## Tests

37 new tests, written before the implementation and watched failing first — `EvalResult` and
`ExpressionEvaluator` unresolved, then `CalculatorKey` unresolved, then the two rewritten sign
tests failing on the old double-sign behaviour.

Coverage on `:library` is filtered by package, so `…currencyedittext.calculator.*` joins
`watchers.*` and `util.*` in the Kover include list. The filtered figure is 94.23%, against the
`minBound(80)` floor.

## Quality gates

`./gradlew build` passes locally. detekt reported `ReturnCount`, `MagicNumber`, `ComplexCondition`
and `TooManyFunctions` along the way; all four are fixed in the code — no `@Suppress`, no baseline
entry, no lowered bound.

`library/api/library.api` is unchanged, as it should be for an all-`internal` addition.

## Not in this pull request

The `PopupWindow` panel, the button, the `calculatorEnabled` attribute and the public setters —
those are PR 2, and they are what will change the ABI.

### DIFF
--- a/library/src/main/java/ru/pravbeseda/currencyedittext/calculator/CalculatorState.kt
+++ b/library/src/main/java/ru/pravbeseda/currencyedittext/calculator/CalculatorState.kt
@@ -28,6 +28,9 @@ private val OPERATOR_KEYS =
 /** Characters after which an operand — and therefore a unary minus — may start. */
 private const val OPERAND_MAY_FOLLOW = "+-*/("
 
+/** The binary operators, and therefore also the characters a pending sign can be written among. */
+private const val OPERATORS = "+-*/"
+
 private fun Char.isOperandChar(): Boolean = isDigit() || this == '.'
 
 private fun Char?.mayPrecedeOperand(): Boolean = this != null && this in OPERAND_MAY_FOLLOW
@@ -77,21 +80,24 @@ internal data class CalculatorState(
             else -> CalculatorState("$expression.")
         }
 
-    private fun appendOperator(operator: Char): CalculatorState =
-        when {
-            // A minus may open an expression or a parenthesis; the others need a left operand.
-            expression.isEmpty() || lastChar == '(' -> {
-                if (operator == '-') CalculatorState(expression + operator) else this
-            }
-
-            lastChar.mayPrecedeOperand() -> {
-                CalculatorState(expression.dropLast(1) + operator)
+    /**
+     * Replaces everything still pending at the end — an operator, and the sign `flipSign` may have
+     * written after it — so that `1*-` plus `+` is `1+` and never `1*+`. Where no operand has
+     * started yet, only a minus is legal, and any other operator leaves the expression alone.
+     */
+    private fun appendOperator(operator: Char): CalculatorState {
+        val pending = expression.takeLastWhile { it in OPERATORS }
+        val settled = expression.dropLast(pending.length)
+        return when {
+            settled.isEmpty() || settled.last() == '(' -> {
+                if (operator == '-') CalculatorState("$settled-") else this
             }
 
             else -> {
-                CalculatorState(expression + operator)
+                CalculatorState(settled + operator)
             }
         }
+    }
 
     private fun appendOpeningParenthesis(): CalculatorState =
         if (expression.isEmpty() || lastChar.mayPrecedeOperand()) {
--- a/library/src/main/java/ru/pravbeseda/currencyedittext/calculator/CalculatorState.kt
+++ b/library/src/main/java/ru/pravbeseda/currencyedittext/calculator/CalculatorState.kt
@@ -1,0 +1,140 @@
+/*
+ * Copyright (c) 2022-2023 Alexander Ivanov
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package ru.pravbeseda.currencyedittext.calculator
+
+import java.math.BigDecimal
+
+private val OPERATOR_KEYS =
+    setOf(
+        CalculatorKey.PLUS,
+        CalculatorKey.MINUS,
+        CalculatorKey.MULTIPLY,
+        CalculatorKey.DIVIDE,
+    )
+
+/** Characters after which an operand — and therefore a unary minus — may start. */
+private const val OPERAND_MAY_FOLLOW = "+-*/("
+
+private fun Char.isOperandChar(): Boolean = isDigit() || this == '.'
+
+private fun Char?.mayPrecedeOperand(): Boolean = this != null && this in OPERAND_MAY_FOLLOW
+
+private fun String.withCharAt(
+    index: Int,
+    char: Char,
+): String = substring(0, index) + char + substring(index + 1)
+
+/**
+ * What the calculator panel has been given so far, as a canonical ASCII expression. Holds no
+ * reference to a view and knows nothing about locale: the panel translates the user's decimal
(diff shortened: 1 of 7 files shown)


library/src/main/java/ru/pravbeseda/currencyedittext/calculator/CalculatorState.kt
```diff
@@ -0,0 +1,140 @@
...
+        }
+
+    private fun appendOperator(operator: Char): CalculatorState =
+        when {
+            // A minus may open an expression or a parenthesis; the others need a left operand.
+            expression.isEmpty() || lastChar == '(' -> {
+                if (operator == '-') CalculatorState(expression + operator) else this
+            }
+
+            lastChar.mayPrecedeOperand() -> {
```
pravbeseda: 🤖 blocking: I am not sure this branch tells a binary operator apart from the unary minus that `flipSign` writes after `*`, `/` or `(`. `OPERAND_MAY_FOLLOW` contains `-`, so a trailing unary minus is treated as "the last operator" and gets swapped for the new key: `press(DIGIT_1, MULTIPLY, SIGN, MULTIPLY)` goes `1` → `1*` → `1*-` → `1**`, and `press(MINUS, PLUS)` goes `-` → `+` — a leading `+`, which the test on line 56 says is not accepted. Both strings come back as `Failure` from the evaluator, and the class doc promises that a press never leaves an expression the user has to back out of. Would the same replacement rule hold if the character being replaced is a unary minus rather than a binary operator?

pravbeseda (commented):
🤖 Reviewed the evaluator, the key state machine, their tests, the Kover filter and the plan document at 1dc8ed1 — 1 blocking finding, 0 suggestions.

library/src/main/java/ru/pravbeseda/currencyedittext/calculator/ExpressionEvaluator.kt
```diff
@@ -0,0 +1,142 @@
...
+    private fun parseProduct(): BigDecimal? {
+        var left = parseFactor()
+        while (left != null && nextIsOneOf("*/")) {
+            val times = input[pos] == '*'
+            pos++
+            val right = parseFactor()
+            left =
+                when {
+                    right == null -> null
+                    times -> left.multiply(right, INTERMEDIATE_MATH)
```
pravbeseda: 🤖 blocking: I am not sure multiplication should use the 16-digit math context here: `9999999999999999*9` evaluates to `89999999999999990.00` rather than the exact `89999999999999991.00`, so the calculator silently changes a valid monetary result. Is that loss of precision intended for multiplication?

library/src/main/java/ru/pravbeseda/currencyedittext/calculator/CalculatorState.kt
```diff
@@ -0,0 +1,140 @@
...
+                CalculatorState(expression.dropLast(1) + operator)
```
pravbeseda: 🤖 blocking: I am not sure an operator following a unary minus is handled here. The key sequence `1`, `×`, `±`, `+`, `2` produces `1*+2`, and `result(2)` returns `Failure`; could this path leave the calculator in an unreadable state after an ordinary operator correction?

pravbeseda (commented):
🤖 Reviewed the calculator evaluator and key-state behavior; 2 blocking findings follow.

library/src/main/java/ru/pravbeseda/currencyedittext/calculator/CalculatorState.kt
pravbeseda: Confirmed and fixed. `dropLast(1)` treated the unary minus `flipSign` writes as if it were the operator to replace, so `1*-` + `×` gave `1**` and `-` + `+` gave a leading `+`, both of which the evaluator rejects — exactly the state the class doc promises never to leave behind.

`appendOperator` now takes the whole pending run of operator characters at the end and replaces it in one go, instead of dropping a single character: `1*-` + `×` is `1*`, `1*-` + `+` is `1+`. Where no operand has started yet — an empty expression, or straight after `(` — only a minus is legal and any other operator leaves the expression untouched, so `-` + `+` stays `-` and `(-` + `+` stays `(-`.

The fix removes a branch rather than adding one: the old "replace the last character" and "append" cases collapse into a single replacement of the pending run.

Two tests cover it, both watched failing first: `an operator replaces a pending sign as well as the operator carrying it` and `an operator cannot turn a leading or bracketed minus into another sign`.

The same defect was reported independently on line 88 by the second reviewer of this round; one fix closes both threads.

library/src/main/java/ru/pravbeseda/currencyedittext/calculator/CalculatorState.kt
pravbeseda: Confirmed and fixed. `1`, `×`, `±`, `+`, `2` did produce `1*+2`, which the evaluator rejects. The cause was in `appendOperator`: it dropped a single trailing character to replace "the last operator", which silently swallowed the unary minus `flipSign` had written after `*`.

`appendOperator` now replaces the whole pending run of operator characters at once, so that sequence gives `1+2`. Where no operand has started yet — an empty expression, or straight after `(` — only a minus is legal and any other operator leaves the expression untouched.

Two tests cover it, both watched failing first: `an operator replaces a pending sign as well as the operator carrying it` and `an operator cannot turn a leading or bracketed minus into another sign`.

The same defect was reported independently on line 87 by the other reviewer of this round; one fix closes both threads.

library/src/main/java/ru/pravbeseda/currencyedittext/calculator/ExpressionEvaluator.kt
pravbeseda: Confirmed and fixed. `9999999999999999*9` did come back as `89999999999999990.00`.

The precision loss was not intended: it was the math context being applied where it is not needed. `BigDecimal` addition, subtraction and multiplication are exact and take no `MathContext` at all — only division has to be told when to stop, or `10 / 3` never terminates. The context is now passed to `divide` alone, so multiplication is exact whatever the operands carry.

`100 / 3 * 3` still ends at 100.00: the division is what keeps the 16 significant digits, and the exact multiplication that follows preserves them.

Test added and watched failing first: `multiplication is exact, however many digits the operands carry`. The KDoc on the constant now says it is division's precision rather than every operation's.

Issue #37, opened 2026-09-07
Sample app: content is drawn under the app bar on Android 15+ (edge-to-edge)


pravbeseda:
## What happens

On a cold start of `:sample` on an API 36 emulator, the `Value` label is completely hidden behind
the app bar and the top half of the `CurrencyEditText` is clipped by it. The field can still be
focused by tapping its lower half, but the label and the field's first line are unreachable.

## Reproduction

```
./gradlew :sample:installDebug
adb shell monkey -p ru.pravbeseda.currencyedittext -c android.intent.category.LAUNCHER 1
adb shell uiautomator dump /sdcard/ui.xml && adb pull /sdcard/ui.xml
```

The dump on an API 36 emulator (1080x2400) reports:

```
[Value]              bounds=[410,53][670,131]
[CurrencyEditText]   bounds=[278,131][803,289]
[app bar title]      bounds=[42,101][450,172]
```

The content starts at y=53, which is inside the status bar, and the app bar occupies roughly
y=90..205 — so it is painted over the first two widgets. No keyboard is involved; this is the
state immediately after launch.

## Cause

`sample/build.gradle:22` sets `targetSdkVersion 37`. From Android 15 (API 35) the platform
enforces edge-to-edge for apps targeting 35 and above, so the window no longer insets its content
for the system bars. `MainActivity` calls `setContentView` at
`sample/src/main/java/.../MainActivity.kt:34` and applies no window insets, and
`sample/src/main/res/layout/activity_main.xml:9` constrains the first `TextView` straight to the
parent top with a 20dp margin.

## Scope

`:sample` only. The published `:library` artifact is unaffected — this is the demo app's window
handling, not the views'.

## Suggested fix

Apply the system-bar insets as padding on the root `ConstraintLayout` in `MainActivity`, the usual
`ViewCompat.setOnApplyWindowInsetsListener` on `binding.root`.

PR #38, opened 2026-09-07
fix(sample): inset the content for the system bars on Android 15+

pravbeseda:
## What

The sample app drew its content under the app bar on Android 15+: the `Value` label was fully
hidden and the top of `CurrencyEditText` was clipped (issue #37).

Reproduced on an API 36 emulator (1080x2400) before the fix:

```
action_bar_container  [0,0][1080,210]     <- painted over the content
textView              [410,53][670,131]
currencyEditText      [278,131][803,289]
```

After:

```
toolbar               [0,0][1080,210]
textView              [410,263][670,341]
currencyEditText      [278,341][803,499]
```

## Why

`:sample` targets SDK 37, and from API 35 the platform enforces edge-to-edge. The decor action bar
of `Theme.AppCompat.Light.DarkActionBar` then stops offsetting the content below itself, so it is
painted straight over the first two widgets.

Note that the fix suggested in the issue — padding the root by the system bar insets alone — is not
enough on its own: that adds 63px (the status bar), while the app bar ends at y=210.

## How

- `styles.xml`: `Theme.AppCompat.Light.DarkActionBar` -> `Theme.AppCompat.Light.NoActionBar`.
- `activity_main.xml`: a vertical `LinearLayout` holding a `statusBarBackground` `View`, an
  `androidx.appcompat.widget.Toolbar` and a `content` `ConstraintLayout` with the previous widgets,
  unchanged.
- `MainActivity`: `setSupportActionBar(binding.toolbar)` plus a window insets listener — the top
  inset becomes the height of `statusBarBackground`, which carries `colorPrimaryDark` so the status
  bar keeps its own tint as it did under the old theme, and the toolbar and the content container
  take the horizontal insets, the container the bottom one as well.

`android:statusBarColor` is not an option here: the platform ignores it for apps targeting SDK 35
and above, so the strip behind the status bar has to be a view the app draws itself.

The two-level layout is what lets the bars stay full-bleed while the title and the content below
them are still inset: padding a single root sideways would cut those backgrounds short.

## Scope

`:sample` only. No change to the published `:library` artifact, and no public API change.

## Testing

`./gradlew build` passes locally (all six gates). Verified by hand on an API 36 emulator, with the
`uiautomator` dumps above, in portrait and in reverse landscape with three-button navigation
(nav bar at x=0..126, toolbar title at x=136); not verified on API 24-34, where the window is not
edge-to-edge and the insets reach the content as zero.

No tests, because the change is confined to the demo app's window handling — there is no logic in
`library/src/main` to cover, and `:sample` has no test source set.

### DIFF
--- a/sample/src/main/java/ru/pravbeseda/currencyedittext/MainActivity.kt
+++ b/sample/src/main/java/ru/pravbeseda/currencyedittext/MainActivity.kt
@@ -20,6 +20,7 @@ import android.os.Bundle
 import androidx.appcompat.app.AppCompatActivity
 import androidx.core.view.ViewCompat
 import androidx.core.view.WindowInsetsCompat
+import androidx.core.view.updateLayoutParams
 import androidx.core.view.updatePadding
 import pravbeseda.databinding.ActivityMainBinding
 import ru.pravbeseda.currencyedittext.CurrencyEditText.Companion.State
@@ -82,8 +83,8 @@ class MainActivity : AppCompatActivity() {
     private fun applySystemBarInsets() {
         ViewCompat.setOnApplyWindowInsetsListener(binding.root) { _, insets ->
             val systemBars = insets.getInsets(WindowInsetsCompat.Type.systemBars())
+            binding.statusBarBackground.updateLayoutParams { height = systemBars.top }
             binding.toolbar.updatePadding(
-                top = systemBars.top,
                 left = systemBars.left,
                 right = systemBars.right,
             )
--- a/sample/src/main/java/ru/pravbeseda/currencyedittext/MainActivity.kt
+++ b/sample/src/main/java/ru/pravbeseda/currencyedittext/MainActivity.kt
@@ -82,8 +82,16 @@ class MainActivity : AppCompatActivity() {
     private fun applySystemBarInsets() {
         ViewCompat.setOnApplyWindowInsetsListener(binding.root) { _, insets ->
             val systemBars = insets.getInsets(WindowInsetsCompat.Type.systemBars())
-            binding.toolbar.updatePadding(top = systemBars.top)
-            binding.root.updatePadding(bottom = systemBars.bottom)
+            binding.toolbar.updatePadding(
+                top = systemBars.top,
+                left = systemBars.left,
+                right = systemBars.right,
+            )
+            binding.content.updatePadding(
+                left = systemBars.left,
+                right = systemBars.right,
+                bottom = systemBars.bottom,
+            )
             insets
         }
     }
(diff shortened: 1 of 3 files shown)


sample/src/main/java/ru/pravbeseda/currencyedittext/MainActivity.kt
```diff
@@ -72,6 +77,17 @@ class MainActivity : AppCompatActivity() {
         }
     }
 
+    // Targeting SDK 35+ makes the window edge-to-edge: without this the system bars
+    // are drawn over the content.
+    private fun applySystemBarInsets() {
+        ViewCompat.setOnApplyWindowInsetsListener(binding.root) { _, insets ->
+            val systemBars = insets.getInsets(WindowInsetsCompat.Type.systemBars())
+            binding.toolbar.updatePadding(top = systemBars.top)
+            binding.root.updatePadding(bottom = systemBars.bottom)
```
pravbeseda: 🤖 blocking: I’m not sure the horizontal system-bar case is handled here: on API 36 with three-button navigation and reverse landscape, the navigation bar occupies x=0..126 while the toolbar title still starts at x=42, so the first part of ‘CurrencyEditText’ is visibly covered. Does ignoring `systemBars.left/right` leave this supported configuration broken?

pravbeseda (commented):
🤖 Reviewed the full PR; 1 blocking finding and 0 suggestions follow.

sample/src/main/java/ru/pravbeseda/currencyedittext/MainActivity.kt
pravbeseda: 🤖 Confirmed on an API 36 emulator, and it is a regression this PR introduced — fixed.

Measured in reverse landscape with three-button navigation (nav bar at x=0..126):

```
b5308b1 (this PR)   toolbar [0,0][2400,189]           title [42,100][326,151]    covered
3f4e23a (base)      action_bar_container [126,0][…]   title [168,100][452,151]   clear
```

The decor action bar applied the left inset itself, so dropping it lost that handling.

Fix: `activity_main.xml` is now a vertical `LinearLayout` holding the `Toolbar` and a `content`
`ConstraintLayout` with the previous widgets. The toolbar takes the top plus horizontal insets, so
its background still spans the full width while the title clears the nav bar; the content container
takes the horizontal plus bottom insets. Re-measured after the fix: toolbar `[0,0][2400,189]`, title
`[136,100][420,151]`, content `[0,189][2400,1080]` — clear of the bar in landscape, and portrait is
unchanged.

PR #40, opened 2026-09-07
feat: opt-in calculator panel for both views

### LINKED ISSUE #39: Instrumented tests do not dex: a backtick test name with spaces is illegal below DEX 040
The `instrumented` job of `ci.yml` has been failing on every push to `main` since 1d27fa4 (#34). It is not a device problem: the build fails at `:library:dexBuilderDebugAndroidTest`, before anything is installed.

## Reproduction

```
./gradlew :library:connectedDebugAndroidTest
```

```
ERROR: .../debugAndroidTest/ru/pravbeseda/currencyedittext/CurrencyEditTextTest.class:
D8: Space characters in SimpleName 'Issue #23 - an empty field reads as zero'
are not allowed prior to DEX version 040
```

## Cause

`library/src/androidTest/java/ru/pravbeseda/currencyedittext/CurrencyEditTextTest.kt:123`

```kotlin
fun `Issue #23 - an empty field reads as zero`() {
```

Backtick method names with spaces compile to class members D8 refuses to dex for `minSdk 24`: spaces in a `SimpleName` need DEX version 040, which is API 30 and above. The same style is fine in `src/test`, which runs on the JVM — this only bites in `src/androidTest`.

## Fix

Rename the instrumented test to a name without spaces, and keep the issue number in it (for example `issue23EmptyFieldReadsAsZero`). The convention holds for every future instrumented test; `src/test` is unaffected and keeps its readable backtick names.

pravbeseda:
Second half of `.github/tasks/calculator-input-panel.md` — the panel, the button and the attribute. PR 1 (#36) added the evaluator and the key state machine.

## What it does

A calculator button at the end of the field, off by default. It opens a panel seeded with the field's current value; `=` writes the result back through `setValue`, so grouping, currency prefix and padding come from `CurrencyInputWatcher` as usual. The system keyboard is not replaced. An expression that is incomplete, divides by zero, or comes out negative in a field with `negativeValueAllow="false"` leaves the field untouched and shows an error state.

`CalculatorPopup` is the only class of the calculator package that touches Android. Key semantics, the arithmetic and the expression line stay in `CalculatorState` and `ExpressionEvaluator`, which `src/test` covers.

## Public API change

Additive only, no existing signature changes:

- `calculatorEnabled` XML attribute on both views (default `false`)
- `setCalculatorEnabled(Boolean)` and `isCalculatorEnabled(): Boolean` on both views
- `CurrencyEditText.onTouchEvent` and `performClick`: overrides of `View` methods, in the ABI dump for that reason. `onTouchEvent` recognises the tap on the compound drawable; `performClick` is what lint's `ClickableViewAccessibility` requires of a view that overrides it.

`updateKotlinAbi` diff is in the same commit.

## Departures from the plan, worth a look

1. `CurrencyMaterialEditText` does **not** delegate `calculatorEnabled` to the inner view, as the plan's step 5 suggested: the layout draws the button through `END_ICON_CUSTOM`, and a delegating setter would make the inner `CurrencyEditText` draw a second one as a compound drawable. It keeps its own flag and claims the end icon only when asked, so a host's own `endIconMode` survives.
2. `CalculatorPopup` is excluded from the Kover filter. It cannot be reached from `src/test`, exactly like the two Views, and an unfiltered figure would report the size of the panel rather than the quality of the tests. Coverage is 88.6% against the unchanged `minBound(80)`.
3. The panel hangs on whichever side of the field has room and caps its height to that room, scrolling inside it. This is the plan's stated risk: in landscape the popup was clipped to the space below the field and showed no keys at all. Verified on the emulator in both orientations.
4. `app:calculatorEnabled="true"` is set on both sample fields — the feature could not be checked by hand otherwise.

## Also here

`fix: rename the instrumented test that cannot be dexed (#39)` — a backtick test name with spaces made `:library:dexBuilderDebugAndroidTest` fail for the whole instrumented suite, so the `instrumented` job of `ci.yml` has been red on `main` since #34. Without it none of the instrumented tests below could run. Fixes #39.

## Tests

- `CalculatorStateTest`, `CalculatorKeyTest`: the negative-result policy, the expression line and the key labels, written before the implementation and watched fail.
- Instrumented tests in both views: `calculatorEnabled` defaults to `false`, and enabling it shows the button. 14/14 green on an emulator.
- `./gradlew build` green locally.
- By hand on the emulator: `1123,48 + 100 =` writes `$ 1.223,48` into the field and fires `onValueChanged`; the panel flips above the field in landscape.

### DIFF
--- a/library/build.gradle
+++ b/library/build.gradle
@@ -82,6 +82,16 @@ kover {
                         'ru.pravbeseda.currencyedittext.util.*'
                 )
             }
+            excludes {
+                // The calculator's panel is the one class of that package that
+                // touches Android, and it is out of the filter for the same
+                // reason as the two Views: src/test cannot reach it, so its
+                // lines would measure the size of the panel rather than the
+                // quality of the tests. Everything it decides — key semantics,
+                // arithmetic, the expression line — lives in CalculatorState
+                // and ExpressionEvaluator, which are measured.
+                classes('ru.pravbeseda.currencyedittext.calculator.CalculatorPopup')
+            }
         }
         variant('debug') {
             // Coverage lands in the build log of every `check`, so the number is
--- a/library/src/main/java/ru/pravbeseda/currencyedittext/CurrencyEditText.kt
+++ b/library/src/main/java/ru/pravbeseda/currencyedittext/CurrencyEditText.kt
@@ -20,7 +20,11 @@ import android.graphics.Rect
 import android.text.InputType
 import android.text.method.DigitsKeyListener
 import android.util.AttributeSet
+import android.view.MotionEvent
+import android.view.View
+import androidx.appcompat.content.res.AppCompatResources
 import androidx.appcompat.widget.AppCompatEditText
+import ru.pravbeseda.currencyedittext.calculator.CalculatorPopup
 import ru.pravbeseda.currencyedittext.model.CurrencyInputWatcherConfig
 import ru.pravbeseda.currencyedittext.util.firstChar
 import ru.pravbeseda.currencyedittext.util.formatMoneyValue
@@ -44,6 +48,7 @@ open class CurrencyEditText(
     private var decimalZerosPadding: Boolean = false
     private var emptyStringForZero: Boolean = true
     private var maxDecimalPlaces: Int
+    private var calculatorEnabled: Boolean = false
 
     private var onValueChanged: OnValueChanged? = null
     private var validator: ((BigDecimal) -> String?)? = null
@@ -93,6 +98,8 @@ open class CurrencyEditText(
                         getBoolean(R.styleable.CurrencyEditText_decimalZerosPadding, false)
                     emptyStringForZero =
                         getBoolean(R.styleable.CurrencyEditText_emptyStringForZero, true)
+                    calculatorEnabled =
+                        getBoolean(R.styleable.CurrencyEditText_calculatorEnabled, false)
                 } finally {
                     recycle()
                 }
@@ -105,6 +112,7 @@ open class CurrencyEditText(
         textWatcher = createTextWatcher()
         this.addTextChangedListener(textWatcher)
         text = this.text // to apply text watcher formatting
+        updateCalculatorButton()
     }
 
     fun setValue(value: BigDecimal) {
@@ -197,6 +205,51 @@ open class CurrencyEditText(
 
     fun isValidState(): Boolean = state == State.OK
 
+    /**
+     * Shows or hides the calculator button. The panel reads the field's configuration when it
+     * opens, so this changes nothing about the watcher and must not invalidate it.
+     */
+    fun setCalculatorEnabled(enabled: Boolean) {
+        calculatorEnabled = enabled
+        updateCalculatorButton()
+    }
+
+    fun isCalculatorEnabled(): Boolean = calculatorEnabled
+
+    /**
+     * Opens the calculator panel under [anchor], which is the field itself unless it sits inside a
+     * [CurrencyMaterialEditText] — there the panel hangs below the whole layout.
+     */
+    internal fun showCalculator(anchor: View) {
+        CalculatorPopup(
+            anchor = anchor,
+            decimalSeparator = getDecimalSeparator(),
+            scale = maxDecimalPlaces,
+            negativeValueAllow = negativeValueAllow,
+            initialValue = getValue(),
+            onAccept = ::setValue,
+        ).show()
+    }
+
+    private fun updateCalculatorButton() {
+        val icon =
+            if (calculatorEnabled) {
+                AppCompatResources.getDrawable(context, R.drawable.currency_edit_text_ic_calculator)
+            } else {
+                null
+            }
+        setCompoundDrawablesRelativeWithIntrinsicBounds(null, null, icon, null)
+    }
+
+    private fun isTouchOnCalculatorButton(x: Float): Boolean {
+        val iconWidth = compoundDrawablesRelative[END_DRAWABLE]?.bounds?.width() ?: return false
+        return if (layoutDirection == LAYOUT_DIRECTION_RTL) {
+            x <= paddingLeft + iconWidth
+        } else {
+            x >= width - paddingRight - iconWidth
+        }
+    }
+
     private fun invalidateTextWatcher() {
         removeTextChangedListener(textWatcher)
         textWatcher = createTextWatcher()
@@ -253,6 +306,22 @@ open class CurrencyEditText(
         super.setText(text, type)
         getText()?.length?.let { setSelection(it) }
     }
+
+    /**
+     * The calculator button is a compound drawable rather than a view of its own, so the tap on it
+     * is recognised here, by where it lands. Everything else is ordinary text editing.
+     */
+    override fun onTouchEvent(event: MotionEvent): Boolean {
+        if (event.action == MotionEvent.ACTION_UP && isTouchOnCalculatorButton(event.x)) {
+            performClick()
+            showCalculator(this)
+            return true
+        }
+        return super.onTouchEvent(event)
+    }
+
+    /** Overridden only because [onTouchEvent] is: an accessibility click has to reach the view. */
+    override fun performClick(): Boolean = super.performClick()
 
     override fun onFocusChanged(
         focused: Boolean,
@@ -314,6 +383,9 @@ open class CurrencyEditText(
     }
 
     companion object {
+        /** Index of the end drawable in `compoundDrawablesRelative`. */
+        private const val END_DRAWABLE = 2
+
         /**
          * Component's state values
          */
--- a/library/src/main/java/ru/pravbeseda/currencyedittext/CurrencyEditText.kt
+++ b/library/src/main/java/ru/pravbeseda/currencyedittext/CurrencyEditText.kt
@@ -17,6 +17,7 @@ package ru.pravbeseda.currencyedittext
 
 import android.content.Context
 import android.graphics.Rect
+import android.graphics.drawable.Drawable
 import android.text.InputType
 import android.text.method.DigitsKeyListener
 import android.util.AttributeSet
@@ -49,6 +50,7 @@ open class CurrencyEditText(
     private var emptyStringForZero: Boolean = true
     private var maxDecimalPlaces: Int
     private var calculatorEnabled: Boolean = false
+    private var calculatorIcon: Drawable? = null
 
     private var onValueChanged: OnValueChanged? = null
     private var validator: ((BigDecimal) -> String?)? = null
@@ -112,7 +114,7 @@ open class CurrencyEditText(
         textWatcher = createTextWatcher()
         this.addTextChangedListener(textWatcher)
         text = this.text // to apply text watcher formatting
-        updateCalculatorButton()
+        if (calculatorEnabled) updateCalculatorButton()
     }
 
     fun setValue(value: BigDecimal) {
@@ -210,6 +212,7 @@ open class CurrencyEditText(
      * opens, so this changes nothing about the watcher and must not invalidate it.
      */
     fun setCalculatorEnabled(enabled: Boolean) {
+        if (calculatorEnabled == enabled) return
         calculatorEnabled = enabled
         updateCalculatorButton()
     }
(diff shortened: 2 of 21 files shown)


library/src/main/java/ru/pravbeseda/currencyedittext/CurrencyEditText.kt
```diff
@@ -197,6 +205,51 @@ open class CurrencyEditText(
...
+    }
+
+    private fun updateCalculatorButton() {
+        val icon =
+            if (calculatorEnabled) {
+                AppCompatResources.getDrawable(context, R.drawable.currency_edit_text_ic_calculator)
+            } else {
+                null
+            }
+        setCompoundDrawablesRelativeWithIntrinsicBounds(null, null, icon, null)
```
pravbeseda: 🤖 blocking: This call sets all four compound drawables, and `init` runs it unconditionally — after `AppCompatEditText` has already applied whatever the host put in XML. Doesn't that mean a `<CurrencyEditText android:drawableStart="@drawable/ic_money" />` in an existing app loses its icon on the next library update, with the calculator still off? I could not find anything that keeps the host's start/top/bottom drawables across this write.

library/src/main/java/ru/pravbeseda/currencyedittext/CurrencyMaterialEditText.kt
```diff
@@ -162,6 +168,25 @@ open class CurrencyMaterialEditText(
...
+     */
+    fun setCalculatorEnabled(enabled: Boolean) {
+        calculatorEnabled = enabled
+        if (enabled) {
+            endIconMode = END_ICON_CUSTOM
+            setEndIconDrawable(R.drawable.currency_edit_text_ic_calculator)
+            setEndIconContentDescription(R.string.currency_edit_text_calculator)
+            setEndIconOnClickListener { editText.showCalculator(this) }
+        } else {
+            endIconMode = END_ICON_NONE
```
pravbeseda: 🤖 blocking: The PR description says the end icon stays the host's until the calculator claims it, and `init` is careful to call the setter only when the attribute is `true`. But this branch runs whether or not the calculator ever owned the icon: a host with `app:endIconMode="clear_text"` that calls `setCalculatorEnabled(false)` — say, from a settings toggle that starts in the off position — ends up with no end icon at all. Is that intended, or should the disable path only touch an icon this view put there?

library/src/main/java/ru/pravbeseda/currencyedittext/CurrencyEditText.kt
```diff
@@ -254,6 +307,22 @@ open class CurrencyEditText(
         getText()?.length?.let { setSelection(it) }
     }
 
+    /**
+     * The calculator button is a compound drawable rather than a view of its own, so the tap on it
+     * is recognised here, by where it lands. Everything else is ordinary text editing.
+     */
+    override fun onTouchEvent(event: MotionEvent): Boolean {
+        if (event.action == MotionEvent.ACTION_UP && isTouchOnCalculatorButton(event.x)) {
```
pravbeseda: 🤖 blocking: `View.dispatchTouchEvent` still calls `onTouchEvent` on a disabled view — the enabled check lives inside `View.onTouchEvent`, which this override returns before reaching. So with `isEnabled = false` and the calculator on, a tap on the icon still opens the panel and `=` still writes into the field through `setValue`. Is a disabled field meant to take a value this way? (The material variant is fine: `TextInputLayout.setEnabled` disables the end icon button itself.)

pravbeseda (commented):
🤖 Reviewed the calculator panel, the button and the attribute at b3796ba: 3 blocking findings, no suggestions.

library/build.gradle
```diff
@@ -82,6 +82,16 @@ kover {
...
             }
+            excludes {
+                // The calculator's panel is the one class of that package that
+                // touches Android, and it is out of the filter for the same
+                // reason as the two Views: src/test cannot reach it, so its
+                // lines would measure the size of the panel rather than the
+                // quality of the tests. Everything it decides — key semantics,
+                // arithmetic, the expression line — lives in CalculatorState
+                // and ExpressionEvaluator, which are measured.
+                classes('ru.pravbeseda.currencyedittext.calculator.CalculatorPopup')
```
pravbeseda: 🤖 blocking: I am not sure this exclusion is compatible with this repository's Definition of Done: it says coverage checks must not be weakened and Android-only logic is not a reason to leave behavior outside `src/test`. The current instrumented tests stop at icon visibility, so opening the popup, wiring keys, accepting a result, rendering an error, dismissal, and placement can all regress while every gate stays green. What currently exercises those paths?

library/src/main/java/ru/pravbeseda/currencyedittext/CurrencyEditText.kt
pravbeseda: 🤖 blocking: Doesn't this erase host-owned compound drawables even when the calculator is left at its default `false`? The constructor always calls `updateCalculatorButton()`, and this four-argument setter overwrites all existing start/top/end/bottom drawables with the supplied values, so an XML `android:drawableStart` (and every other slot) disappears as soon as the view is created; toggling the calculator also discards those drawables.

library/src/main/java/ru/pravbeseda/currencyedittext/CurrencyEditText.kt
pravbeseda: 🤖 blocking: This branch checks only whether some end drawable occupies the hit area, not whether the calculator is enabled or the view is enabled. After a host installs its own end drawable programmatically, a calculator-off field opens the panel when that icon is tapped; a disabled field with the calculator icon also opens it because this code runs before `super.onTouchEvent`. Should either of those paths be allowed to invoke an opt-in calculator?

library/src/main/java/ru/pravbeseda/currencyedittext/CurrencyEditText.kt
```diff
@@ -254,6 +307,22 @@ open class CurrencyEditText(
...
+            performClick()
+            showCalculator(this)
+            return true
+        }
+        return super.onTouchEvent(event)
+    }
+
+    /** Overridden only because [onTouchEvent] is: an accessibility click has to reach the view. */
+    override fun performClick(): Boolean = super.performClick()
```
pravbeseda: 🤖 blocking: An accessibility `ACTION_CLICK` or keyboard click enters `performClick()` directly, but this override only delegates to `super` and never opens the panel. Since the compound drawable is not a separate accessibility node, doesn't this make the calculator on `CurrencyEditText` reachable only by a pointer tap on the icon?

library/src/main/java/ru/pravbeseda/currencyedittext/CurrencyMaterialEditText.kt
pravbeseda: 🤖 blocking: I am not sure the host's end-icon state survives a runtime toggle here. For a field configured with `END_ICON_CLEAR_TEXT` (or a host custom icon/listener), enabling the calculator replaces that state and disabling it forces `END_ICON_NONE`, leaving the previous mode, drawable, description, and listener lost even though the calculator no longer owns the slot. Doesn't that make `setCalculatorEnabled(true); setCalculatorEnabled(false)` destructive to existing `TextInputLayout` behavior?

pravbeseda (commented):
🤖 Reviewed the calculator panel integration at b3796ba; 5 blocking findings and 0 suggestions follow.

library/src/main/java/ru/pravbeseda/currencyedittext/CurrencyEditText.kt
pravbeseda: Confirmed, and fixed. Two changes: `init` now calls `updateCalculatorButton()` only when the calculator is actually on, so a field left at the default writes no drawables at all; and the write itself keeps the other three slots, taking only the end one.

`theHostsOwnDrawablesSurviveTheCalculatorButton` covers it: a host `drawableStart` survives enabling and disabling the calculator, and the end slot is empty once the calculator is off.

Both reviewers of this round raised this; the other thread carries the same fix.

pravbeseda: Confirmed, and fixed. Two changes: `init` now calls `updateCalculatorButton()` only when the calculator is actually on, so a field left at the default writes no drawables at all; and the write itself keeps the other three slots, taking only the end one.

`theHostsOwnDrawablesSurviveTheCalculatorButton` covers it: a host `drawableStart` survives enabling and disabling the calculator, and the end slot is empty once the calculator is off.

Both reviewers of this round raised this; the other thread carries the same fix.

library/src/main/java/ru/pravbeseda/currencyedittext/CurrencyMaterialEditText.kt
pravbeseda: Confirmed, and fixed. `setCalculatorEnabled` now returns early when the state does not change, so switching off a calculator that was never on touches no icon at all — a host's `END_ICON_CLEAR_TEXT` survives it. The disable path also clears the click listener it installed, so nothing of the calculator's is left behind on the slot it gives up.

`disablingACalculatorTheLayoutNeverHadKeepsTheHostsEndIcon` covers it.

What the fix deliberately does not do is restore a host icon the calculator displaced while it was on: enabling the calculator claims the slot, and a host that wants its own icon back sets it. Remembering and restoring the previous mode, drawable, description and listener is state for a case nobody has, and it is the case where the two owners of one slot genuinely conflict.

Both reviewers of this round raised this; the other thread carries the same fix.

pravbeseda: Confirmed, and fixed. `setCalculatorEnabled` now returns early when the state does not change, so switching off a calculator that was never on touches no icon at all — a host's `END_ICON_CLEAR_TEXT` survives it. The disable path also clears the click listener it installed, so nothing of the calculator's is left behind on the slot it gives up.

`disablingACalculatorTheLayoutNeverHadKeepsTheHostsEndIcon` covers it.

What the fix deliberately does not do is restore a host icon the calculator displaced while it was on: enabling the calculator claims the slot, and a host that wants its own icon back sets it. Remembering and restoring the previous mode, drawable, description and listener is state for a case nobody has, and it is the case where the two owners of one slot genuinely conflict.

Both reviewers of this round raised this; the other thread carries the same fix.

library/src/main/java/ru/pravbeseda/currencyedittext/CurrencyEditText.kt
pravbeseda: Confirmed, and fixed. The decision now lives in one place, `opensCalculatorOn(event)`, which requires the view to be enabled, the calculator to be on, and the tap to land on the icon; `onTouchEvent` asks it and otherwise hands the event to `super` untouched. The hit test also matches against the icon this view installed rather than whatever occupies the end slot, so a host's own end drawable no longer opens an opt-in panel.

`onlyAnEnabledFieldWithTheCalculatorOnOpensThePanel` covers all three: on and enabled opens, `isEnabled = false` does not, and a host end drawable with the calculator off does not.

Both reviewers of this round raised this; the other thread carries the same fix.

pravbeseda: Confirmed, and fixed. The decision now lives in one place, `opensCalculatorOn(event)`, which requires the view to be enabled, the calculator to be on, and the tap to land on the icon; `onTouchEvent` asks it and otherwise hands the event to `super` untouched. The hit test also matches against the icon this view installed rather than whatever occupies the end slot, so a host's own end drawable no longer opens an opt-in panel.

`onlyAnEnabledFieldWithTheCalculatorOnOpensThePanel` covers all three: on and enabled opens, `isEnabled = false` does not, and a host end drawable with the calculator off does not.

Both reviewers of this round raised this; the other thread carries the same fix.

library/build.gradle
pravbeseda: Taken in the direction the repository's own rule points: logic that `src/test` cannot reach is a signal to move it out, not a licence to skip the test. The placement arithmetic — which side the panel hangs on and what height it may take — is now `CalculatorPlacement.choose(spaceBelow, spaceAbove, wanted)`, a pure function of three `Int`s in the calculator package, covered by `CalculatorPlacementTest`: fits below, flips above, capped to the roomier side, and the tie. That is where this PR's landscape bug actually was, and it is now inside the measured code rather than outside it.

What stays excluded is `CalculatorPopup` itself, and only that class: after the move it inflates the layout, wires each button to a `CalculatorKey`, and calls `PopupWindow`. It cannot be reached from `src/test` for the same reason the two Views cannot — the module has no Robolectric, and adding one to measure inflation is a larger change than this PR. The exclusion is not a lowered bound: `minBound(80)` is untouched and the filtered figure is 88.8%.

The paths you list are covered where they are decidable: key semantics and the expression line in `CalculatorStateTest`, arithmetic and rounding in `ExpressionEvaluatorTest`, the accept-and-refuse policy in `CalculatorState.result`, placement in the new test, and the button's visibility in both instrumented suites. Inflation, click wiring and dismissal are checked by hand on an emulator, which the PR description states.

library/src/main/java/ru/pravbeseda/currencyedittext/CurrencyEditText.kt
pravbeseda: The observation is right, the change it points to is not, and we are keeping the behaviour as written — with the limitation now documented in the README.

The fix the comment implies — opening the panel from `performClick()` — would hijack every ordinary click on the field: `performClick` is the click on the whole `EditText`, not on the icon. The alternative, a custom accessibility action, was weighed and turned down here: it puts "Open calculator" in the node's action menu rather than giving the icon the real button a screen reader expects, so it adds code without reaching parity with the other component, and it leaves the exact scenario in this comment — an accessibility `ACTION_CLICK` — behaving as it does today anyway.

The accessible path already exists and is the one the README now names: `CurrencyMaterialEditText` renders the button through `END_ICON_CUSTOM`, a real focusable button with a content description. `CurrencyEditText` draws a compound drawable by design, which is what makes the calculator opt-in on a view that must otherwise stay an ordinary `EditText`.

Ground: the written approach is equally valid for what this component is, and the proposed one only adds code.

PR #41, opened 2026-09-07
chore: pin the library's Kotlin language and api version to 2.1

pravbeseda:
## What

`:library` now compiles with `languageVersion = 2.1` and `apiVersion = 2.1`.

## Why

The Kotlin metadata version stamped into the published classes follows the language
version. Built with the compiler's own 2.4, `CurrencyEditText.class` carries
`mv=[2,4,0]`, and a consumer on Kotlin 2.1 or 2.2 gets *"class file was compiled by a
newer Kotlin compiler"*. With the pin the same class carries `mv=[2,1,0]`, readable by
every compiler from 2.1 up. `apiVersion` keeps the source honest about it: no stdlib
member newer than 2.1 can slip into the artifact.

Verified on the bytecode, `javap -v` on the release variant:

| | metadata version |
|---|---|
| before | `mv=[2,4,0]` |
| after | `mv=[2,1,0]` |

## Consumer-visible change

This widens what can consume the artifact; it does not narrow it. `library/api/library.api`
is unchanged — the metadata version is not part of the ABI dump — so no `updateKotlinAbi`
is needed.

## Trade-off

`kotlinc` prints `Language version 2.1 is deprecated and its support will be removed in a
future version of Kotlin`. It fails nothing: `warningsAsErrors` is configured for Android
Lint only, and `allWarningsAsErrors` is not set for Kotlin. The next major Kotlin line will
force the choice again.

## Tests

No tests, because this is a build-script change — the exception CLAUDE.md names alongside
pure renames and doc edits. The effect is verified by inspecting the emitted metadata
version rather than by a unit test, which cannot observe it.

## Checks

- `./gradlew build` — green
- `./gradlew :library:checkKotlinAbi` — green, dump unchanged
- Kover: 88.79% against the floor of 80

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,8 +32,10 @@ implementation 'ru.pravbeseda:CurrencyEditText:<insert-latest-version-here>'
 
 Requires `minSdk 24` (Android 7.0). Releases up to 1.1.1 support `minSdk 19`.
 
-The artifact is compiled against Kotlin 2.1, so it can be consumed by Kotlin 2.1 and
-newer. Release 2.0.0 was built with Kotlin 2.4 metadata and needs a 2.4 compiler.
+The artifact carries Kotlin 2.1 metadata and declares `kotlin-stdlib:2.1.0`, so a project
+on Kotlin 2.1 or newer can consume it with nothing special in its own build; a newer
+consumer keeps its own stdlib. Release 2.0.0 carries Kotlin 2.4 metadata and needs a 2.4
+compiler.
 
 For versions, kindly head over to
 the [releases page](https://github.com/pravbeseda/CurrencyEditText/releases)
--- a/README.md
+++ b/README.md
@@ -32,6 +32,9 @@ implementation 'ru.pravbeseda:CurrencyEditText:<insert-latest-version-here>'
 
 Requires `minSdk 24` (Android 7.0). Releases up to 1.1.1 support `minSdk 19`.
 
+The artifact is compiled against Kotlin 2.1, so it can be consumed by Kotlin 2.1 and
+newer. Release 2.0.0 was built with Kotlin 2.4 metadata and needs a 2.4 compiler.
+
 For versions, kindly head over to
 the [releases page](https://github.com/pravbeseda/CurrencyEditText/releases)
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -43,6 +43,14 @@ Updates arrive through Renovate, configured in `renovate.json5`: weekly, patch r
 automerged, minor and major read by a person, androidx grouped, the toolchain never automerged.
 The file only configures the bot — the GitHub App itself is installed on the repository by hand.
 
+The Kotlin language and api versions are pinned to 2.1 in `library/build.gradle`, and
+`kotlin.stdlib.default.dependency=false` in `gradle.properties` hands the stdlib version to
+the catalog's `kotlinStdlib = "2.1.0"`, declared by both modules. Together they are what
+lets a consumer on Kotlin 2.1 use the published artifact: the pin alone stamps `mv=[2,1,0]`
+on the library's own classes, but the POM would still carry the plugin's `kotlin-stdlib`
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -32,6 +32,9 @@ implementation 'ru.pravbeseda:CurrencyEditText:<insert-latest-version-here>'
 
 Requires `minSdk 24` (Android 7.0). Releases up to 1.1.1 support `minSdk 19`.
 
+The artifact is compiled against Kotlin 2.1, so it can be consumed by Kotlin 2.1 and
```
pravbeseda: 🤖 blocking: I am not sure this sentence holds for a consumer that does nothing special in its own build. The pin does stamp `mv=[2,1,0]` on the classes (checked with `javap -v` on the release variant of this branch), but the published POM and Gradle module metadata still declare `org.jetbrains.kotlin:kotlin-stdlib:2.4.10` at compile scope — the version follows the Kotlin Gradle plugin, not `languageVersion` — and that jar carries `mv=[2,4,0]` of its own.

I built a scratch Android library module on Kotlin 2.1.20 / AGP 8.13.0 against this branch's artifact from `mavenLocal()`. Gradle resolved `kotlin-stdlib:2.1.20 -> 2.4.10`, and `compileReleaseKotlin` failed with:

```
e: kotlin-stdlib-2.4.10.jar!/META-INF/kotlin-stdlib.kotlin_module Module was compiled with an incompatible version of Kotlin. The binary version of its metadata is 2.4.0, expected version is 2.1.0.
e: Consumer.kt:8:19 Unresolved reference 'listOf'.
```

That is the same error the PR sets out to remove, now coming from the stdlib instead of from `CurrencyEditText.class`. With `kotlin-stdlib` forced back to 2.1.20 in the consumer's build the same module compiles, so the library's own metadata is fine; it is the transitive stdlib that still keeps a 2.1 or 2.2 consumer out. Was the intent that the artifact be consumable by Kotlin 2.1 as-is, the way this sentence and the PR's "widens what can consume the artifact" read — or is a consumer below 2.4 expected to deal with the stdlib version on its side?

pravbeseda (commented):
🤖 Reviewed the whole PR at fd27868 — the language/api version pin in `library/build.gradle` and the README note — and verified the emitted metadata and a Kotlin 2.1 consumer build against the branch's artifact: 1 blocking finding, 0 suggestions.

README.md
pravbeseda: Confirmed, and fixed. Both halves of the claim reproduce here: the generated POM declared `org.jetbrains.kotlin:kotlin-stdlib:2.4.10` at compile scope, and `CollectionsKt` in that jar carries `mv=[2,4,0]`. The pin on its own only moved the wall.

The version came from the Kotlin Gradle plugin, which adds the stdlib at its own version unless told not to. So the fix is to declare it ourselves:

- `kotlin.stdlib.default.dependency=false` in `gradle.properties`
- `kotlinStdlib = "2.1.0"` in the version catalog, declared by `:library` on `api` (the configuration the plugin used, so the POM keeps compile scope) and by `:sample` on `implementation`, since the flag is project-wide

The published POM now declares `kotlin-stdlib:2.1.0`. That is a floor, not a ceiling: a consumer on a newer Kotlin brings its own stdlib and Gradle's conflict resolution keeps it.

Verified on the scratch consumer this comment describes — Kotlin 2.1.20, AGP 8.13.0, artifact from `mavenLocal()`, same source file both times:

| artifact | POM stdlib | `compileReleaseKotlin` |
|---|---|---|
| before the fix | 2.4.10 | 4 errors, `metadata is 2.4.0, expected version is 2.1.0` |
| after the fix | 2.1.0 | BUILD SUCCESSFUL, resolved `kotlin-stdlib:2.1.20` |

`apiVersion = 2.1` is what makes declaring the older stdlib safe: no member newer than 2.1 can reach the bytecode.

The README sentence you commented on has been rewritten to say what is now true, and the two halves are recorded together in CLAUDE.md, since changing either one alone re-breaks the other.

PR #42, opened 2026-09-08
Calculator panel: a public style contract, and placement after the keyboard closes

pravbeseda:
Two defects of the calculator panel, both reported from a Material3 app: the panel opened clipped to a few key rows, and it painted itself in the platform's default purple instead of the app's colours.

## What changed

**Placement.** `CalculatorPopup` measured `getWindowVisibleDisplayFrame()` while the soft keyboard was still up, and the popup — `focusable = true` — then closed that keyboard itself. The panel was placed against a screen that no longer existed: flipped above the field and capped to the room the keyboard had left, so its scroll view showed about two rows. It now asks the keyboard to close first and places the panel once the window frame has answered, with a timeout for the case where no keyboard was open.

**Colours.** The keys used `?android:attr/borderlessButtonStyle` and the background `?android:attr/colorBackground`, which a Material3 theme does not fill in. Material attributes were the first fix and were reversed: `?attr/colorSurface` and `?attr/colorOnSurface` are declared only by Material, and under an AppCompat theme — the one `:sample` itself uses — inflating the panel throws `InflateException`. Instead the panel now reads five colour roles from a style the host points `currencyCalculatorStyle` at, over defaults that resolve under any theme.

## Public API change

New public XML attributes: `currencyCalculatorStyle`, plus the five roles of `<declare-styleable name=\"CurrencyCalculator\">` — `calculatorPanelBackgroundColor`, `calculatorExpressionTextColor`, `calculatorErrorTextColor`, `calculatorKeyTextColor`, `calculatorOperatorTextColor`. A host names the roles it wants and leaves the rest, on the app theme or on the `android:theme` overlay of a single field. Documented in README.

`library/src/main/res-public/values/public.xml` was an empty `<resources/>` and produced no `public.txt` in the AAR, so every library resource was public. It now names the sixteen XML attributes, which makes the panel's layout, drawables, colours and styles private to the artifact. No Kotlin API changed, so `library/api/library.api` is untouched.

## How it was verified

- `./gradlew build` green after `clean`, coverage 81.15% against the 80 floor.
- `:library:connectedAndroidTest` — 23/23 on API 26 and API 36.
- Red before green: `CalculatorKeyRoleTest` (no `isOperator` to compile against), `thePanelInflatesUnderAnAppCompatTheme` (`InflateException`), `aHostStyleRepaintsTheRolesItNamesAndLeavesTheRest` (partial overrides lost four roles until the defaults were read in their own pass — `defStyleRes` is consulted only when `defStyleAttr` resolves to nothing).
- On an API 26 emulator, before the placement fix the panel opened above the field with two of six rows visible and the keyboard still up; after it, the keyboard closes and the panel opens below the field at full height.

No test for the keyboard wait itself: it is IME timing inside a `PopupWindow`, which `src/test` cannot reach and `src/androidTest` has no Espresso or activity rules for. It was verified by hand on the emulator, before and after.

`CalculatorColors` reads the host theme through a `Context`, so it joined `CalculatorPopup` in the Kover exclusion list for the same stated reason; the coverage bound itself is untouched.

Plan and rulings: `.github/tasks/calculator-panel-placement-and-theme.md`.

### DIFF
--- a/library/build.gradle
+++ b/library/build.gradle
@@ -101,7 +101,13 @@ kover {
                 // quality of the tests. Everything it decides — key semantics,
                 // arithmetic, the expression line — lives in CalculatorState
                 // and ExpressionEvaluator, which are measured.
-                classes('ru.pravbeseda.currencyedittext.calculator.CalculatorPopup')
+                // CalculatorColors is out for the same reason: it reads the
+                // host's theme through a Context, which src/test has none of.
+                // The instrumented CalculatorPanelThemeTest covers it instead.
+                classes(
+                        'ru.pravbeseda.currencyedittext.calculator.CalculatorPopup',
+                        'ru.pravbeseda.currencyedittext.calculator.CalculatorColors'
+                )
             }
         }
         variant('debug') {
--- a/.github/tasks/calculator-panel-placement-and-theme.md
+++ b/.github/tasks/calculator-panel-placement-and-theme.md
@@ -1,0 +1,61 @@
+# Calculator panel: placement after the keyboard closes, and theming
+
+## Goal
+
+The calculator panel opens clipped to a few key rows and paints itself in the platform's default
+purple instead of the host app's Material theme. Both are library bugs, reproduced on a Material3
+dark-themed app (see the issue screenshot): fix the placement race and make the panel take its
+colors from the app's theme.
+
+## The two defects
+
+1. **Stale measurement.** `CalculatorPopup.showAnchored` (`CalculatorPopup.kt:104-118`) reads
+   `getWindowVisibleDisplayFrame()` while the soft keyboard is still open, so `spaceBelow` is small,
+   `CalculatorPlacement.choose` flips the panel above the field and caps its height to `spaceAbove`.
+   The panel is then shown with `PopupWindow(..., focusable = true)`, which takes focus and closes
+   the IME — the screen frees up, and the panel keeps hanging at the height measured for a screen
+   that no longer exists. Inside the `ScrollView` that shows as ~2.5 visible rows.
+2. **Framework theme attributes.** Every key uses `style="?android:attr/borderlessButtonStyle"`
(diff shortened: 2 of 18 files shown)


library/build.gradle
```diff
@@ -101,7 +101,13 @@ kover {
                 // quality of the tests. Everything it decides — key semantics,
                 // arithmetic, the expression line — lives in CalculatorState
                 // and ExpressionEvaluator, which are measured.
-                classes('ru.pravbeseda.currencyedittext.calculator.CalculatorPopup')
+                // CalculatorColors is out for the same reason: it reads the
+                // host's theme through a Context, which src/test has none of.
+                // The instrumented CalculatorPanelThemeTest covers it instead.
+                classes(
```
pravbeseda: 🤖 suggestion: with `CalculatorColors` joining the exclusion list, the sentence this comment opens with — the panel is "the one class of that package that touches Android" — no longer holds, and `CLAUDE.md` (Architecture, the `calculator/` bullet) still says `CalculatorPopup` is "the only one excluded from the Kover filter". Is the intent that the exclusion list is the single place the reason lives, with the two sentences brought in line so a reader does not meet one statement here and a different one in `CLAUDE.md`?

pravbeseda (commented):
🤖 Reviewed 0dcb357 — the keyboard-aware placement, the per-field panel guard, the style contract and the tests: 0 blocking findings, 1 suggestion.

library/build.gradle
pravbeseda: Correct, and fixed. The comment opened by calling the panel the one class of the package that touches Android and then, four lines down, excluded a second one — and `CLAUDE.md` repeated the first half. Both now say the same thing: the two classes that touch Android are out of the filter, with the reason kept in `library/build.gradle` and `CLAUDE.md` pointing at it rather than restating it. The exclusion list and the coverage bound are unchanged.

